PR #10, opened 2017-11-08
Use Fastly!

mrasu:
Use Fastly!

### DIFF
--- a/src/wovnio/wovnphp/Store.php
+++ b/src/wovnio/wovnphp/Store.php
@@ -47,7 +47,7 @@
         'url_pattern_name' => 'query',
         'url_pattern_reg' => '((\?.*&)|\?)wovn=(?P<lang>[^&]+)(&|$)',
         'query' => array(),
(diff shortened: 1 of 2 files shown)


src/wovnio/wovnphp/Store.php
```diff
@@ -47,7 +47,7 @@ private function defaultSettings() {
         'url_pattern_name' => 'query',
         'url_pattern_reg' => '((\?.*&)|\?)wovn=(?P<lang>[^&]+)(&|$)',
         'query' => array(),
-        'api_url' => 'https://api.wovn.io/v0/',
+        'api_url' => 'http://wovn.global.ssl.fastly.net/v0/',
```
hythof: Looks good.
Why don't you use `https` instead of `http`?

mrasu: done

mrasu:
@hythof ping

PR #12, opened 2017-11-16
Not send wovn-ignore or script tag to Translation API 

mrasu:
Not send wovn-ignore or script tag to Translation API to cache faster and not send secure information.

* Add snippet to ensure always snippet is inserted.
* Remove wovn-ignore element and children
* Remove Form to avoid CSRF
* Remove Script to avoid CSRF
* Remove content surrounded by "backend-wovn-ignore" comment


src/wovnio/html/HtmlConverter.php
```diff
@@ -0,0 +1,155 @@
...
+   *
+   * @return array converted html and HtmlReplaceMarker
+   */
+  public function convertToAppropriateForApiBody() {
+    // encoding detection uses 30% of execution time for this method, should we make encoding as setting?
+    $encoding = mb_detect_encoding($this->html, array('UTF-8', 'EUC-JP', 'SJIS', 'eucJP-win', 'SJIS-win', 'JIS', 'ISO-2022-JP', 'ASCII'));
+    $dom = str_get_html($this->html, false, false, $encoding, false);
+    $marker = new HtmlReplaceMarker();
+    $this->insertSnippet($dom);
+    $this->removeWovnIgnore($dom, $marker);
```
a-bougouin: For Unified Values support, I am not sure we can completely remove `wovn-ignore` content.
@kyuden can you confirm.

src/wovnio/html/HtmlConverter.php
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * Convert to appropriate HTML to send to Translation API
+   * e.g.) remove wovn-ignore content
+   *
+   * @return array converted html and HtmlReplaceMarker
+   */
+  public function convertToAppropriateForApiBody() {
+    // encoding detection uses 30% of execution time for this method, should we make encoding as setting?
+    $encoding = mb_detect_encoding($this->html, array('UTF-8', 'EUC-JP', 'SJIS', 'eucJP-win', 'SJIS-win', 'JIS', 'ISO-2022-JP', 'ASCII'));
```
a-bougouin: I agree. If user always use same encoding, they should be able to specify settings for faster execution. I user does not specify encoding, we should keep detection.

If we do that, we should also document special setup for optimization purpose (not related bit need to do same with curl library).

test/html/HtmlReplaceMakerTest.php
```diff
@@ -0,0 +1,51 @@
+<?php
+require_once 'src/wovnio/html/HtmlReplaceMarker.php';
+
+use Wovnio\Html\HtmlReplaceMarker;
+
+class HtmlReplaceMarkerTest extends PHPUnit_Framework_TestCase {
```
a-bougouin: Please make tests with at least 11 keys.
I don't see the problem here, but first wovnphp version had revert issues with keys ending by `1` and `11` (or `2` and `20`, etc.). We need to secure that case for future changes.

src/wovnio/wovnphp/API.php
```diff
@@ -19,18 +20,22 @@ public static function url($store, $headers, $original_content) {
     public static function translate($store, $headers, $original_content) {
       $translated_content = NULL;
       $api_url = self::url($store, $headers, $original_content);
+
+      $token = $store->settings['project_token'];
+      $converter = new HtmlConverter($original_content, $token);
+      list($converted_html, $maker) = $converter->convertToAppropriateForApiBody();
```
a-bougouin: Typo: `$maker` -> `$marker`

a-bougouin:
> Add snippet to ensure always snippet is inserted

Was it requested? When to expect API to not return HTML with snippet?

If there is a bug with API not returning HTML with snippet, then we need to fix the API.
I am concerned that adding the snippet will make it more difficult to catch error and user may lose backend benefit (SEO) without noticing in a short window of time.

src/wovnio/html/HtmlConverter.php
mrasu: @kyuden please check

src/wovnio/html/HtmlConverter.php
mrasu: done. `encoding` option is created

test/html/HtmlReplaceMakerTest.php
mrasu: done!

src/wovnio/wovnphp/API.php
mrasu: done!

mrasu:
@a-bougouin 
For when WOVN.php cannot reach translation API.

In that case, "j.wovn.io" is still working because "j.wovn.io" doesn't depend on API.

src/wovnio/html/HtmlConverter.php
a-bougouin: thanks

test/html/HtmlReplaceMakerTest.php
a-bougouin: thanks

a-bougouin:
@mrasu, then could we insert the snippet after API request?
If snippet exists, do nothing. If snippet does not exists, then add snippet and trigger a Rollbar error. I want us to be notified somehow that snippet was not inserted, so that we can investigate and fix the problem.

What do you think?

mrasu:
@a-bougouin seems nice. I added "data-wovnio-type" to snippet

mrasu:
@a-bougouin  :+1:

src/wovnio/html/HtmlConverter.php
kyuden: In current implementation of swapper, wovn-ignore is needed. original content is key to find the destination.

kyuden: I talk with @mrasu in person. we don't need wovn-ignore content. but wovn-ignore tag is needed.

mrasu: ok done.

a-bougouin:
Tests need to be fixed.

@WOVNio/developers can someone review please.

test/APITest.php
```diff
@@ -129,19 +140,54 @@ public function testTranslateWithCustomLangAliases() {
       $result = API::translate($store, $headers, $html);
       $this->assertEquals($expected_result, $result);
     }
+/*
```
a-bougouin: Why must comment test?

mrasu: thanks, removed

PR #15, opened 2017-11-29
If the request is for the default lang, do not send an API request

jsand-minimal:
- This is an option that is off by default

### Comments
Rough draft, no tests, and might need refactoring

### DIFF
--- a/src/wovn_interceptor.php
+++ b/src/wovn_interceptor.php
@@ -28,7 +28,12 @@
     $headers->responseOut();
 
     if(!empty($buffer) && $buffer != strip_tags($buffer)) {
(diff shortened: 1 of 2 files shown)


src/wovn_interceptor.php
```diff
@@ -28,7 +28,12 @@
     $headers->responseOut();
 
     if(!empty($buffer) && $buffer != strip_tags($buffer)) {
-      $translated_buffer = API::translate($store, $headers, $buffer);
+      if ($store->settings['disable_api_on_default_lang'] && $headers->lang() == $store->settings['default_lang']) {
+        $translated_buffer = $buffer;
```
mrasu: at least, we need to add
* hreflang
* snippet

PR #16, opened 2017-11-30
Feature/disable api request for default lang

jsand-minimal:
When the request is for the default language, don't do an API request, but still add the snippet and hreflang tags.

~Still needs more tests and some refactoring.~

Tests added

~Branched off of feature/remove_wovn_ignore_to_translation_api, but if that branch is merged we can fix this branch and merge it into master.~

Update: Rebased on master

### DIFF
--- a/src/wovnio/html/HtmlConverter.php
+++ b/src/wovnio/html/HtmlConverter.php
@@ -48,7 +48,7 @@ class HtmlConverter
    *
    * @return array converted html and HtmlReplaceMarker
    */
-  public function convertToAppropriateForApiBody($removeParts=true)
+  public function convertToAppropriateForApiBody($removeParts = true)
   {
     if ($this->encoding && in_array($this->encoding, self::$supported_encodings)) {
       $encoding = $this->encoding;
@@ -65,7 +65,6 @@ class HtmlConverter
 
     $marker = new HtmlReplaceMarker();
     if ($removeParts) {
-      echo 'doing it here';
       $this->removeWovnIgnore($dom, $marker);
       $this->removeForm($dom, $marker);
       $this->removeScript($dom, $marker);
@@ -79,7 +78,6 @@ class HtmlConverter
     unset($dom);
 
     if ($removeParts) {
-      echo 'doing it';
       $converted_html = $this->removeBackendWovnIgnoreComment($converted_html, $marker);
     }
     return array($converted_html, $marker);
--- a/src/wovnio/html/HtmlConverter.php
+++ b/src/wovnio/html/HtmlConverter.php
@@ -1,29 +1,45 @@
 <?php
+
 namespace Wovnio\Html;
 
+require_once 'src/wovnio/wovnphp/Url.php';
+require_once 'src/wovnio/wovnphp/Lang.php';
+
+use Wovnio\Wovnphp\Url;
+use Wovnio\Wovnphp\Lang;
 use Wovnio\ModifiedVendor\simple_html_dom;
+
 /**
  * Convert html via Simple HTML DOM Parser
  *
  * @see http://simplehtmldom.sourceforge.net/manual.htm
  */
-class HtmlConverter {
+class HtmlConverter
+{
   public static $supported_encodings = array('UTF-8', 'EUC-JP', 'SJIS', 'eucJP-win', 'SJIS-win', 'JIS', 'ISO-2022-JP', 'ASCII');
 
   private $html;
   private $encoding;
   private $token;
+  private $store;
+  private $headers;
+
 
   /**
    * HtmlConverter constructor.
    * @param String $html
    * @param String $encoding
    * @param String $token project_token
+   * @param Store $store
+   * @param Headers $headers
    */
-  public function __construct($html, $encoding, $token) {
+  public function __construct($html, $encoding, $token, $store = null, $headers = null)
+  {
     $this->html = $html;
     $this->encoding = $encoding;
     $this->token = $token;
+    $this->store = $store;
+    $this->headers = $headers;
   }
 
   /**
@@ -32,7 +48,8 @@ class HtmlConverter {
    *
    * @return array converted html and HtmlReplaceMarker
    */
-  public function convertToAppropriateForApiBody() {
+  public function convertToAppropriateForApiBody()
+  {
     if ($this->encoding && in_array($this->encoding, self::$supported_encodings)) {
       $encoding = $this->encoding;
     } else {
@@ -43,6 +60,9 @@ class HtmlConverter {
     $dom = simple_html_dom::str_get_html($this->html, $encoding, false, false, $encoding, false);
     $marker = new HtmlReplaceMarker();
     $this->insertSnippet($dom);
+    if (isset($this->store) && isset($this->headers)) {
+      $this->insertHreflangTags($dom);
+    }
     $this->removeWovnIgnore($dom, $marker);
     $this->removeForm($dom, $marker);
     $this->removeScript($dom, $marker);
@@ -64,7 +84,8 @@ class HtmlConverter {
    *
    * @param simple_html_dom_node $dom
    */
-  private function insertSnippet($dom) {
+  private function insertSnippet($dom)
+  {
     foreach ($dom->find('script') as $node) {
       if (strpos($node->getAttribute('src'), '//j.wovn.io/') !== false ||
         strpos($node->getAttribute('src'), '//j.dev-wovn.io:3000/') !== false) {
@@ -74,7 +95,7 @@ class HtmlConverter {
 
     $token = $this->token;
     $insert_tags = array('head', 'body', 'html');
-    foreach($insert_tags as $tag_name) {
+    foreach ($insert_tags as $tag_name) {
       $parents = $dom->find($tag_name);
       if (count($parents) > 0) {
         $parent = $parents[0];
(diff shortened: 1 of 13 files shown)


src/wovnio/html/HtmlConverter.php
```diff
@@ -74,6 +82,50 @@ private function insertSnippet($dom) {
...
+
+    $insert_tags = array('head', 'body', 'html');
+    foreach($insert_tags as $tag_name) {
+      $parents = $dom->find($tag_name);
+      if (count($parents) > 0) {
+        $parent = $parents[0];
+        $hreflangTags = array('');
+        foreach ($lang_codes as $lang_code) {
+          $href = Url::addLangCode($this->headers->url, $this->store, $lang_code, $this->headers);
+          array_push($hreflangTags, '<link rel="alternate" hreflang="' . $lang_code . '" href="' . $href . '">');
```
a-bougouin: `$lang_code` needs to follow ISO 639 specification in `hreflang` attribute:
```
array_push($hreflangTags, '<link rel="alternate" hreflang="' . Lang.iso639_1Normalization($lang_code) . '" href="' . $href . '">');
```

test/html/HtmlConverterTest.php
```diff
@@ -1,14 +1,77 @@
...
+    $env = array();
+    $file = parse_ini_file(dirname(__FILE__) . '/../mock_env' . $num . '.ini');
+    $env = $file['env'];
+    return $env;
+  }
+
+  /**
+    * @group new
+   */
+  public function testHreflang() {
```
a-bougouin: We need to add tests to check that chinese hreflang values are correctly set with ISO 639 spec.

src/wovnio/html/HtmlConverter.php
jsand-minimal: Thanks, we also need to make sure that we are using any custom lang aliases defined. I'll fix this.

a-bougouin: yes, that's right

test/html/HtmlConverterTest.php
a-bougouin: Needs tests with custom lang aliases too.

src/wovnio/html/HtmlConverter.php
jsand-minimal: Updated, please check

src/wovnio/wovnphp/Lang.php
```diff
@@ -110,7 +110,7 @@ public static function getEnglishNamesArray() {
      * @return String The ISO639-1 code of the language.
      */
     public static function iso639_1Normalization($lang_code) {
-      return str_replace('zh-CHT', 'zh-Hant', str_replace('zh-CHS', 'zh-Hans', $lang_code));
+      return Lang::$lang[$lang_code]['iso639-1'];
```
suplo: should check `Lang::$lang[$lang_code]` exists or not

test/html/HtmlConverterTest.php
```diff
@@ -1,14 +1,77 @@
...
+    /* $headers = new Headers($env, $store); */
+
+    $converter = new HtmlConverter($html, 'UTF-8', $token, $store, $headers);
+    list($translated_html, $marker) = $converter->convertToAppropriateForApiBody();
+
+    $expected_html_text = file_get_contents('test/fixtures/real_html/stack_overflow_hreflang_expected.html');
+
+    $this->assertEquals($expected_html_text, $translated_html);
+    $this->assertEquals(1, 1);
+    return;
```
suplo: the code from here to end of function could be removed

test/html/HtmlConverterTest.php
suplo: testInsertHrefLang might be a more proper name

src/wovnio/html/HtmlConverter.php
```diff
@@ -74,6 +82,49 @@ private function insertSnippet($dom) {
...
+   *
+   * @param simple_html_dom_node $dom
+   */
+  public function insertHreflangTags($dom) {
+    $lang_codes = $this->store->settings['supported_langs'];
+    foreach ($dom->find('link') as $node) {
+      $hreflangValue = $node->getAttribute('hreflang');
+      if (in_array(Lang::getCode($hreflangValue), $lang_codes)) {
+        // remove node
+        $node->outertext = '';
```
hythof: I think that following code is more clear.
```
$node->parentNode->removeChild($node);
```

src/wovnio/html/HtmlConverter.php
```diff
@@ -74,6 +82,49 @@ private function insertSnippet($dom) {
...
+    foreach($insert_tags as $tag_name) {
+      $parents = $dom->find($tag_name);
+      if (count($parents) > 0) {
+        $parent = $parents[0];
+        $hreflangTags = array('');
+        foreach ($lang_codes as $lang_code) {
+          $href = Url::addLangCode($this->headers->url, $this->store, $lang_code, $this->headers);
+          array_push($hreflangTags, '<link rel="alternate" hreflang="' . Lang::iso639_1Normalization($lang_code) . '" href="' . $href . '">');
+        }
+        $parent->innertext = implode("\n", $hreflangTags) . "\n" . $parent->innertext;
```
hythof: I think that it is better using `appendChild` instead of `innertext`.

src/wovnio/html/HtmlConverter.php
```diff
@@ -74,6 +82,49 @@ private function insertSnippet($dom) {
 
     $token = $this->token;
     $insert_tags = array('head', 'body', 'html');
+    foreach($insert_tags as $tag_name) {
```
a-bougouin: isn't it duplicate of line 111?

src/wovnio/html/HtmlConverter.php
jsand-minimal: There is no appendChild. Also, we are copying the above function that Hiroaki wrote.
http://simplehtmldom.sourceforge.net/manual_api.htm

src/wovnio/html/HtmlConverter.php
jsand-minimal: There's no removeChild
http://simplehtmldom.sourceforge.net/manual_api.htm

src/wovnio/html/HtmlConverter.php
jsand-minimal: fixed, please check

src/wovnio/html/HtmlConverter.php
```diff
@@ -84,6 +100,37 @@ private function insertSnippet($dom) {
     }
   }
 
+  /**
+   * Insert hreflang tags for all supported_langs
+   *
+   * @param simple_html_dom_node $dom
+   */
+  public function insertHreflangTags($dom) {
+    $lang_codes = $this->store->settings['supported_langs'];
```
suplo: $lang_codes should be `$store->settings['supported_langs']` + `iso639_1Normalization` lang code (for Chinese) imo

src/wovnio/html/HtmlConverter.php
hythof: Okay.

src/wovnio/html/HtmlConverter.php
hythof: Okay.

src/wovnio/wovnphp/Lang.php
```diff
@@ -1,116 +1,126 @@
...
+   * Provides the ISO639-1 code for a given lang code.
+   * Source: https://support.google.com/webmasters/answer/189077?hl=en
+   *
+   * @param String $lang_code Code of the language.
+   *
+   * @return String The ISO639-1 code of the language.
+   */
+  public static function iso639_1Normalization($lang_code)
+  {
+    if (isset(Lang::$lang[$lang_code])) {
```
a-bougouin: I don't know if it is possible, but if user put custom lang alias in supported languages, wouldn't the custom lang alias be sent to this function?
If so, shouldn't we also resolve lang code with custom lang aliases before getting iso 639 code?

toinoudivad: we shouldn't use the custom lang alias for the href langs, only for the url

src/wovnio/html/HtmlConverter.php
```diff
@@ -74,6 +82,49 @@ private function insertSnippet($dom) {
 
     $token = $this->token;
     $insert_tags = array('head', 'body', 'html');
+    foreach($insert_tags as $tag_name) {
+      $parents = $dom->find($tag_name);
+      if (count($parents) > 0) {
+        $parent = $parents[0];
+        $parent->innertext = "<script src=\"//j.wovn.io/1\" data-wovnio=\"key=$token\" data-wovnio-type=\"backend_without_api\" async></script>" . $parent->innertext;
```
mrasu: j.wovn.io is inserted at 132 as same.

test/APITest.php
```diff
@@ -198,7 +198,7 @@ public function testTranslateWithErrorHandled() {
              $this->equalTo($expected_url),
              $this->equalTo($expected_data),
              $this->equalTo(1.0)
-           )
+           )L
```
suplo: plz remove `L`

test/html/HtmlConverterTest.php
```diff
@@ -1,14 +1,24 @@
 <?php
 require_once 'src/wovnio/html/HtmlConverter.php';
 require_once 'src/wovnio/html/HtmlReplaceMarker.php';
-
+require_once 'src/wovnio/wovnphp/Utils.php';
+require_once 'src/wovnio/wovnphp/Headers.php';
 require_once 'src/wovnio/modified_vendor/simple_html_dom.php';
 
 use Wovnio\Html\HtmlConverter;
+use Wovnio\Wovnphp\Url;
```
suplo: it's strange that we have no `require_once` for Url but still able to use it.

test/APITest.php
suplo: fixed.

test/html/HtmlConverterTest.php
a-bougouin: Probably because it is required in another file that is require here

suplo: Yes, you're right. I tried remove all require_once and tests still pass

src/wovnio/html/HtmlConverter.php
```diff
@@ -84,6 +100,37 @@ private function insertSnippet($dom) {
...
+  /**
+   * Insert hreflang tags for all supported_langs
+   *
+   * @param simple_html_dom_node $dom
+   */
+  public function insertHreflangTags($dom) {
+    $lang_codes = $this->store->settings['supported_langs'];
+    foreach ($dom->find('link') as $node) {
+      $hreflangValue = $node->getAttribute('hreflang');
+      if (in_array(Lang::getCode($hreflangValue), $lang_codes)) {
```
a-bougouin: I think here `$lang_code` should be iso normalized too

src/wovnio/wovnphp/Lang.php
a-bougouin: Yes, I wan't a check here to convert custom lang code into wovn lang code for normalization.
@suplo, any plan to do that before merging?

jsand-minimal: The previous version of this function didn't handle that case. From looking at the code, lang_code is always the wovn lang code, if it's custom it is specified.

I can modify to to support this case if you'd like. What do you think?

src/wovnio/html/HtmlConverter.php
jsand-minimal: I modified Lang::getCode to support Normalized values

src/wovnio/wovnphp/Lang.php
a-bougouin: Maybe ok from now, because previous version was like that.
I think we should make a task to either:
- support that case
- make sure user put wovn lang code in supported languages, not custom lang codes

jsand-minimal:
Looks good to me, thank you so much, Dat!

Shojigo:
or wovn.php in the HtmlConverter.php file
line 5 and 6
require_once ‘src/wovnio/wovnphp/Url.php’;
require_once ‘src/wovnio/wovnphp/Lang.php’;

our user saying there seem to have error on this 

the right code should be this 

require_once ‘WOVN.php/src/wovnio/wovnphp/Url.php’;
require_once ‘WOVN.php/src/wovnio/wovnphp/Lang.php’;

PR #17, opened 2017-12-04
Call requestOut before intercepting buffer

suplo:

Calling requestOut before intercepting buffer to keep `_ENV` as same as visiting page without wovn intercepted.


src/wovn_interceptor.php
```diff
@@ -25,6 +25,8 @@
   list($store, $headers) = Utils::getStoreAndHeaders($_SERVER);
 
   $_ENV['WOVN_TARGET_LANG'] = $headers->lang();
+  $headers->requestOut();
```
hythof: Looks good call in here.
But even in line 33 called `requestOut()`, is it necessary at line 33?.

src/wovnio/wovnphp/Headers.php
```diff
@@ -16,7 +16,7 @@ class Headers {
     public $redisUrl;
     // PHP ONLY
     public $maskedRequestURI;
-
```
hythof: 👍

src/wovn_interceptor.php
toinoudivad: line 33 is responseOut and is different method

suplo: line 33 is `responseOut`, not `requestOut`

test/HeadersTest.php
```diff
@@ -885,11 +872,41 @@ public function testRequestOutWithUseProxyFalse () {
...
+    $store->settings['url_pattern_name'] = 'query';
+    $store->settings['url_pattern_reg'] = '((\?.*&)|\?)wovn=(?P<lang>[^&]+)(&|$)';
+
+    $env = $this->getEnv('_query');
+    $_SERVER['REQUEST_URI'] = $env['REQUEST_URI'];
+
+    $headers = new Headers($env, $store);
+    $headers->requestOut();
+    $this->assertEquals('', $env['QUERY_STRING']);
+    $this->assertEquals('/t.php?wovn=ja', $env['REQUEST_URI']);
```
toinoudivad: why is the language still inside the request uri ?

test/HeadersTest.php
```diff
@@ -885,11 +872,41 @@ public function testRequestOutWithUseProxyFalse () {
...
+    $store->settings['url_pattern_name'] = 'path';
+    $store->settings['url_pattern_reg'] = '\/(?P<lang>[^\/.]+)(\/|\?|$)';
+
+    $env = $this->getEnv('_path');
+    $_SERVER['REQUEST_URI'] = $env['REQUEST_URI'];
+
+    $headers = new Headers($env, $store);
+    $headers->requestOut();
+    $this->assertEquals('/t.php?', $env['REQUEST_URI']);
+    $this->assertEquals('/t.php', $env['REDIRECT_URL']);
```
toinoudivad: can we check it's here before the requestOut ?

test/HeadersTest.php
suplo: the code for `query` is commented out

```
            // removed this in case of redirect the lang is lost
            //$this->_env['REQUEST_URI'] = $this->removeLang($this->_env['REQUEST_URI']);
```

This is intention IMO, so I did not uncomment code

test/HeadersTest.php
suplo: Sure. I added.

test/HeadersTest.php
toinoudivad: I guess it should not affect user's site behavior so it might be fine

test/HeadersTest.php
toinoudivad: thanks

src/wovn_interceptor.php
hythof: I mistook it.
Thanks.

toinoudivad:
What's the status, are we merging this?

suplo:
I will merge

PR #20, opened 2017-12-14
Stop using simple_html_dom because of memory leak issue

suplo:

- Stop using simple_html_dom because of memory leak issue
- Do not send request to API if current lang is default lang
- Inject snippet/hreflangs using regex (doing)

### DIFF
--- a/src/wovnio/html/HtmlConverter.php
+++ b/src/wovnio/html/HtmlConverter.php
@@ -4,7 +4,6 @@ namespace Wovnio\Html;
 
 use Wovnio\Wovnphp\Url;
 use Wovnio\Wovnphp\Lang;
-use Wovnio\ModifiedVendor\simple_html_dom;
 
 /**
  * Convert html via Simple HTML DOM Parser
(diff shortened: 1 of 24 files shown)


src/wovnio/html/HtmlConverter.php
```diff
@@ -48,64 +48,39 @@ public function __construct($html, $encoding, $token, $store = null, $headers =
...
     $token = $this->token;
-    $insert_tags = array('head', 'body', 'html');
-    foreach ($insert_tags as $tag_name) {
-      $parents = $dom->find($tag_name);
-      if (count($parents) > 0) {
-        $parent = $parents[0];
-        $parent->innertext = "<script src='//j.wovn.io/1' data-wovnio='key=$token' data-wovnio-type='backend_without_api' async></script>" . $parent->innertext;
-        return;
+    $snippet_code = "<script src='//j.wovn.io/1' data-wovnio='key=$token' data-wovnio-type='backend_without_api' async></script>";
+    $parent_tags = array('<head>', '<body>', '<html>');
```
mrasu: any tag can have some attributes so we should allow

suplo: You're right. I changed code to cover that. Thanks

src/wovnio/html/HtmlConverter.php
```diff
@@ -45,64 +44,39 @@ public function __construct($html, $encoding, $token, $store = null, $headers =
...
-    return array($converted_html, $marker);
+    return array($this->html);
   }
 
   /**
    * Insert wovn's snippet to ensure snippet is always inserted.
    * When snippet is always inserted, do nothing
    *
-   * @param simple_html_dom_node $dom
+   * @param simple_html_dom_node $html
```
toinoudivad: param is not simple_html_dom_node anymore I think, right?

suplo: You're right. I will change. Thank you

src/wovnio/html/HtmlConverter.php
```diff
@@ -45,64 +44,39 @@ public function __construct($html, $encoding, $token, $store = null, $headers =
...
 
     $token = $this->token;
-    $insert_tags = array('head', 'body', 'html');
-    foreach ($insert_tags as $tag_name) {
-      $parents = $dom->find($tag_name);
-      if (count($parents) > 0) {
-        $parent = $parents[0];
-        $parent->innertext = "<script src='//j.wovn.io/1' data-wovnio='key=$token' data-wovnio-type='backend_without_api' async></script>" . $parent->innertext;
-        return;
+    $snippet_code = "<script src='//j.wovn.io/1' data-wovnio='key=$token' data-wovnio-type='backend_without_api' async></script>";
```
toinoudivad: the snippet code should have backend=true in the data-wovnio attribute as well as default lang and current lang, you can refer to @jsand-minimal PR here : https://github.com/WOVNio/WOVN.php/pull/18

suplo: Thank you. I will change

suplo: Done. Please check again

src/wovnio/html/HtmlConverter.php
suplo: Done

src/wovnio/html/HtmlConverter.php
```diff
@@ -45,98 +44,89 @@ public function __construct($html, $encoding, $token, $store = null, $headers =
...
    */
-  private function insertSnippet($dom)
+  private function insertSnippet($html)
   {
-    foreach ($dom->find('script') as $node) {
-      if (strpos($node->getAttribute('src'), '//j.wovn.io/') !== false ||
-        strpos($node->getAttribute('src'), '//j.dev-wovn.io:3000/') !== false) {
-        return;
-      }
+    $snippet_regex = "<\<script(.*)src='(j.wovn.io|j.dev-wovn.io)'(.*)><\/script>";
```
toinoudivad: I'm sorry I approved already but I think we should always insert the snippet even if existing and in that case remove the existing one, what do you think?

suplo: Good idea. Will make change soon. Thank you

jsand-minimal: Does this regex match the standard WOVN.io snippet in your environment? From looking at it I think it's wrong, I'm working on a correct version.
sample snippet:
`<script src="//j.wovn.io/1" data-wovnio="key=hS7IIB" async></script>`

jsand-minimal: This regex should match all environments of WOVN, and also it matches the standard widget.

`$snippet_regex = "<script(.*)src=[^ ]*(j.[^.]*wovn.io)(.*)><\/script>";`

src/wovnio/html/HtmlConverter.php
```diff
@@ -45,98 +44,89 @@ public function __construct($html, $encoding, $token, $store = null, $headers =
...
   }
 
+  private function buildSnippetCode()
+  {
+    $token = $this->token;
+    $current_lang = $this->headers->lang();
+    $default_lang = $this->store->settings['default_lang'];
+    $url_pattern = $this->store->settings['url_pattern_name'];
+    $lang_code_aliases_json = json_encode($this->store->settings['custom_lang_aliases']);
+//    $snippet_code = "<script src='//j.wovn.io/1' data-wovnio='key=$token' data-wovnio-type='backend_without_api' async></script>";
```
jsand-minimal: This is the old incorrect one, so we can remove this comment now, right?

suplo: Yes. I will remove

jsand-minimal:
Closed #18 because this PR contains the fix

src/wovnio/html/HtmlConverter.php
suplo: Code updated. Please check again

suplo: Thank you. I changed your regex a bit. Please check again.

src/wovnio/html/HtmlConverter.php
suplo: Removed. Thanks

PR #23, opened 2018-01-10
remove dunit

mrasu:
because dunit doesn't become error when test failed

### Comments
at https://circleci.com/gh/WOVNio/WOVN.php/92
you can see `dunit` is passed even when phpunit fails

docker's command is the same with dunit


mrasu:
by seeing repositories of famous php framework, they're using travis.
it might be better for this repository too.

toinoudivad:
I'm fine with removing but dunit's purpose originally was to check that the syntax was correct for different versions of php. (for example using a function not available in php 5.3 would make dunit fail)

mrasu:
@toinoudivad 
Yes! it's useful to check syntax for each version!
So, I didn't change the behavior of dunit.
This PR changes only not use dunit but do the same thing in separate commands.

at hack/run_test.sh, the script checks syntax and run phpunit.

PR #27, opened 2018-01-15
Modify simple_html_dom to fit our style

mrasu:
Modify simple_html_dom to fit our style

### Comments
Change is

* Separate simple_html_dom.php to SimpleHtmlDom.php and SimpleHtmlDomNode.php
* Change indentation to space
* Stop using `define` but use class' constant

There is no change on logic


mrasu:
@suplo I did it!

PR #28, opened 2018-01-15
Stop depending on real `wovn.ini` at test

mrasu:
Stop depending on real `wovn.ini` at test

### Comments
Currently, tests assume there is default configuration or there is no file.
But there is in reality.

To divide dependency,

1. Change Store's constructor to get parsed configuration as array
2. Make simple Builder to load and parse configuration.
3. Fix tests

### DIFF
--- a/src/wovnio/wovnphp/Store.php
+++ b/src/wovnio/wovnphp/Store.php
@@ -13,22 +13,30 @@
     private $values = null;
 
     /**
-     *  Constructor of the Store class
-     *  
-     *  @param string $settingsFile A settings file name 
-     *  @return void
+     * @param string $settingFileName
+     * @return Store
      */
-    public function __construct($settingsFile='') {
-      if ($settingsFile==='') {
-        $settingsFile = DIRNAME(__FILE__) . '/../../../../wovn.ini';
-      }
-      $defaultSettings = $this->defaultSettings();
-      if (file_exists($settingsFile)) {
-        $userSettings = $this->updateSettings(parse_ini_file($settingsFile, true));
+    public static function createFromFile($settingFileName) {
+      if (file_exists($settingFileName)) {
+        $userSettings = parse_ini_file($settingFileName, true);
       }
       else {
-        $userSettings = array();
+        $userSettings = null;
       }
+
+      return new Store($userSettings);
+    }
+
+    /**
+     *  Constructor of the Store class
+     *  
+     *  @param array $userSettings
+     *  @return void
+     */
+    public function __construct($userSettings) {
+      $defaultSettings = $this->defaultSettings();
+      $userSettings = $this->updateSettings($userSettings);
+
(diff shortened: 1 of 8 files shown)


src/wovnio/wovnphp/Store.php
```diff
@@ -79,6 +87,10 @@ private function defaultSettings() {
      * @return array The new settings of the user
      */
     public function updateSettings($vals=array()) {
+      if (isset($vals['default_lang']) === false) {
+        return array();
```
a-bougouin: Why reset all setting if the default language is not present?
Is it new from this PR?

mrasu: This is existing behavior,
https://github.com/WOVNio/WOVN.php/pull/28/files#diff-949fcf6c77696a804a40a883c4af4753L26

before:
If there isn't a file, use array().
If there is no "default_lang", error happens

a-bougouin: > If there is no "default_lang", error happens

I don't see that in the previous code?

mrasu: at `$vals['default_lang']`.
php raise error when the value is not in dictionary(array)

a-bougouin: Then, can we merge setting with default ones as done at `https://github.com/WOVNio/WOVN.php/blob/master/src/wovnio/wovnphp/Store.php#L32`.
Or we could just prevent the error at `updateSettings()`.

That way we avoid error and don't reset settings.

mrasu: What is the problem of resetting setting when a setting is invalid?
I just make almost the same behavior with now.

If the problem is about current behavior, this is out of the scope of this PR.

suplo:
Test `1) APITest::testTranslateWithErrorHandled` fails for me even with/without `wovn.ini`. 
I don't have this problem with `master` branch.

Could you please check?

mrasu:
@a-bougouin @suplo ping

src/wovnio/wovnphp/Store.php
a-bougouin: I disagree with resetting every option if one is missing.
I agree with setting default values for missing "required" options, but not resetting all.

mrasu: My question is " the problem is about current behavior" or not.
If your disagreement is about current behavior, this should be discussed in different place.
Because we didn't discuss anything about how to handle invalid configuration (for me, raising error is better but there are other options).

a-bougouin: If you want error, then why is this change needed?
I don't see code raising explicit error. If not, then this change is not needed and may lead to error later. So, can't we just discard this change and let the line `$vals['default_lang'] = Lang::getCode($vals['default_lang']);` raise error as now?

mrasu: ok done

PR #31, opened 2018-03-16
Remove SimpleHtmlDom MAX_FILE_SIZE limit and ensure WOVN.php does not

a-bougouin:
Remove SimpleHtmlDom MAX_FILE_SIZE limit and ensure WOVN.php does not crash if SimpleHtmlDom::str_get_html returns a boolean instead of an object


suplo:
Personally, I think it's better if we give warning if $html size is over some limit.

PR #38, opened 2018-06-28
Check Content-Type of response if exists.

mrasu:
Check Content-Type of response if exists.

### Comments
Stop translation when Content-Type is not html-ish.

Note:
Because when Content-Type is not set, Apache add extra header based on file extension, consider html when the header is not set.

### DIFF
--- a/src/wovnio/wovnphp/Utils.php
+++ b/src/wovnio/wovnphp/Utils.php
@@ -43,7 +43,7 @@ class Utils {
         continue;
       }
       if (strtolower(trim($header_data[0])) == 'content-type') {
-        if (!preg_match('/\/html/', $header_data[1])) {
+        if (!preg_match('/html/', $header_data[1])) {
           return false;
         }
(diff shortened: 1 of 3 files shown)


src/wovnio/wovnphp/Utils.php
```diff
@@ -34,7 +34,21 @@ public static function changeHeaders($buffer, $store) {
...
+  public static function isHtml($response_headers, $buffer) {
+    // Because Apache sets header automatically when Content-Type is not set,
+    //    check the header only when it exists.
+    foreach($response_headers as $response_header) {
+      $header_data = explode(':', $response_header);
+      if (count($header_data) < 2) {
+        continue;
+      }
+      if (strtolower(trim($header_data[0])) == 'content-type') {
+        if (!preg_match('/\/html/', $header_data[1])) {
```
hythof: It's need care for xhtml.

PR #39, opened 2018-07-03
Feature/ignore paths and regex

toinoudivad:
Feature to let users ignore some paths and globs
Users can set ignore paths in wovn.ini file as follows:
```
ignore_paths[] = "/images/"
ignore_paths[] = "/assets/"
```
Users can set ignore regex in wovn.ini as follows:
```
ignore_regex[] = "/globs\/images$/i" 
```


src/wovnio/wovnphp/Utils.php
```diff
@@ -53,11 +53,33 @@ public static function isAmp($buffer) {
...
         preg_match(self::AUDIO_FILE_PATTERN, $uri) ||
         preg_match(self::VIDEO_FILE_PATTERN, $uri) ||
-        preg_match(self::DOC_FILE_PATTERN, $uri));
+        preg_match(self::DOC_FILE_PATTERN, $uri) ||
+        self::checkIgnorePaths($uri, $store) ||
+        self::checkIgnoreGlobs($uri, $store));
+  }
+
+  private static function checkIgnorePaths($uri, $store = null) {
+    if (null === $store || !isset($store->settings['ignore_paths']) || isset($store->settings['ignore_paths']) && empty($store->settings['ignore_paths'])) return false;
```
hythof: The check logic is to hard for reading.
How about implement helper to `Store`?

But this is a little thing, so I approved it.

src/wovnio/wovnphp/Store.php
```diff
@@ -79,6 +79,8 @@ private function defaultSettings() {
         'wovn_dev_mode' => false,
         'use_server_error_settings' => false,
         'disable_api_request_for_default_lang' => false,
+        'ignore_paths' => array(),
+        'ignore_globs' => array(),
```
a-bougouin: Down in the code, it is using `preg_match`, so should we rename this setting `ignore_regex` instead?

toinoudivad: we use ignore globs on wovnrb so I thought it made sense to use globs here

src/wovnio/wovnphp/Utils.php
toinoudivad: in Store I already implement security check to set array as empty, do you think I could avoid extra checks here and only check if setting is empty only? this means I assume settings is always here

src/wovnio/wovnphp/Store.php
a-bougouin: I agree we should have unified settings across backends, but if we use this setting as a regular expression user may input actual GLOB pattern that won't work.

toinoudivad: ok I will use regex you are right

a-bougouin: thanks

src/wovnio/wovnphp/Utils.php
hythof: If there is like `Store->ignorePaths()`, I think that it's helpful for read and for use.
For example, following codes is easy to read for me.
```
# Store.php
function ignorePaths() {
  if (isset($this->settings['ignore_paths'])) {
    $this->settings['ignore_paths'];
  } else {
    return array();
  }
}

# Utils.php
if (null === $store) {
  return false;
}
foreach($store->ignorePaths() as $path) {
    if (strpos($uri, $path) !== false) {
      return true;
    }
  }
  return false;
}
```

a-bougouin:
This branch has conflicts

mrasu:
What is the status of this PR?

Shojigo:
can i have this deployed next monday?
thanks

toinoudivad:
@Shojigo Just merged so new WOVN.php will have this feature

Shojigo:
thanks mate!

PR #40, opened 2018-07-04
Fix wovn_index_sample.php to not include query parameters during filepath lookup

a-bougouin:
### Purpose/goal of PR
Fix `wovn_index_sample.php` to not include query parameters during filepath lookup

### Comments
Because `$_SERVER['REQUEST_URI']` may contain query parameters, looking for filepath to include may fail and will fallback to displaying content of index file instead of actual file.


suplo:
Could we have a test for this change?

a-bougouin:
@suplo I agree, I added

PR #41, opened 2018-07-12
Custom ignore class

shubhakarreddy:
`Custom ignore class` setting

### DIFF
--- a/src/wovnio/html/HtmlConverter.php
+++ b/src/wovnio/html/HtmlConverter.php
@@ -283,7 +283,7 @@ class HtmlConverter
     $class_attr = $node->getAttribute('class');
     if ($class_attr) {
       $classes_to_ignore = $this->store->settings['ignore_class'];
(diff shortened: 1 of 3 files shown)


src/wovnio/html/HtmlConverter.php
```diff
@@ -278,6 +279,18 @@ function _removeWovnIgnore($node, $marker) {
     }
   }
 
+  function _removeCustomIgnoreClass($node, $marker) {
+    $class_attr = $node->getAttribute('class');
+    if ($class_attr) {
+      $classes_to_ignore = $this->store->settings['ignore_class'];
+      $classes = array_filter(explode(' ', $class_attr));
```
a-bougouin: Do you think it is enough to split by space? Maybe we should split by whitespace?

shubhakarreddy: I made the changes, thanks.

a-bougouin: thanks

PR #42, opened 2018-08-13
Add PHP evaluation with SSI support

a-bougouin:
### Purpose/goal of PR
Evaluate PHP with SSI suppor

### Comments
`file_get_contents` which is used for SSI support does not evaluate PHP
code within `<?php` and `?>`. This commit uses `include` instead of
`file_get_contents`, which evaluate PHP code.

### DIFF
--- a/src/wovnio/wovnphp/SSI.php
+++ b/src/wovnio/wovnphp/SSI.php
@@ -11,7 +11,7 @@
     public static function readFileRecursive($includePath, $rootDir, &$limit) {
       $ssi_include_regexp = '/<!--#include virtual="(.+?)"\s*-->/';
       $includeDir = dirname($includePath);
-      $code = file_get_contents($includePath);
+      $code = self::get_contents($includePath);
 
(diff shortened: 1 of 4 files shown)


src/wovnio/wovnphp/SSI.php
```diff
@@ -38,4 +38,13 @@ public static function readFileRecursive($includePath, $rootDir, &$limit) {
 
       return $code;
     }
+
+    private static function get_contents($includePath) {
+      ob_start();
+      include $includePath;
+      $contents = ob_get_contents();
```
hythof: Also you can use `ob_get_clean()`.
This is `ob_get_contents()` + `ob_end_clean()`.
http://phpspot.net/php/man/php/function.ob-get-clean.html

hythof:
@a-bougouin This changing contains to disable SSI on the default.
I agree with you because it's safer and faster in common case.
I think that to discuss to cs-team(shoji) is better before merging.

a-bougouin:
@hiroshi I will

PR #44, opened 2018-08-22
Switch to CircleCI 2

suplo:
### Comments

I committed 'composer.lock' to the repository. Is there any specific reason it should not be in the repository?


toinoudivad:
Just need to update the branch with current master

PR #45, opened 2018-08-23
hotfix/redirection_to_index_instead_of_404

tim-wovn:
The main issue is the following:
If the user entered the URL for a file or folder that did not exist, 
the code would search the root directory for an index file.
Now it only searches the url given and yields
a 404 error if it can't find an appropriate index file there.

### To summarize the changes in this PR:

1. In the original code, we ran a security check.  We test that the base URL (e.g. www.wovn.io) is in the final resolved URL (e.g. `www.wovn.io/../../etc/`).  If not, we were returning the base URL.  In the new code, we return a 404 error.

2. If the URL cannot be parsed by parse_url() (e.g. `www.wovn.io//////////////../../etc`), a 404 page will be shown.  Originally, the code showed the website's main page. 

3. An existing inconsistency was found with realpath() behaviour between PHP and HHVM in resolving '..' in paths.  So I implemented functions for resolving '.', '..', and '/'.  We still use realpath() but only after we've cleaned that paths first in order to avoid the inconsistent behaviour.

4. We search for a 404 index file in the website's root directory in the sample index file.  If customers wish to use a 404 file with a custom name or in a custom location, they will need to ensure that they are loading the right 404 file.


tim-wovn:
Needs tests.

tim-wovn:
Also, I don't think their 404 page is actually being triggered.

wovn_index_sample.php
```diff
@@ -20,5 +20,12 @@
...
   header($_SERVER["SERVER_PROTOCOL"] . " 404 Not Found");
-  echo "Page Not Found";
+
+  # Look for 404.html file in the root directory
+  $paths_404 = [dirname(__FILE__) . "/404.html", ];
+  $included_404 = wovn_helper_include_by_paths_with_ssi($paths_404);
+
+  if (!$included_404) {
+    echo "Page Not Found";
+  }
```
tim-wovn: @hythof @a-bougouin Is this a reasonable thing to do?

tim-wovn: (Please view the file diff--the 4 line preview does not show all of the logic that I'm concerned about.)

src/wovn_helper.php
```diff
@@ -3,11 +3,29 @@
 require_once(__DIR__ . '/wovnio/wovnphp/SSI.php');
 use Wovnio\Wovnphp\SSI;
 
+function remove_dots_from_path($path) {
```
a-bougouin: This should probably be private

test/WovnIndexSampleTest.php
```diff
@@ -46,9 +46,12 @@ public function testDetectMultipleFiles () {
...
   }
 
-  public function testInvalidPath () {
+  /**
+   * @runInSeparateProcess
+   */
+  public function testInvalidPathBlip () {
     $this->touch('index.php');
-    $this->assertEquals('This is index.php', $this->runWovnIndex('/../../index.php'));
+    $this->assertEquals('Page Not Found', $this->runWovnIndex('/../../index.php'));
```
tim-wovn: /../../index.php should be an invalid directory when appended to root, right?

Also, this fails unless @runInSeparateProcess is appended???

wovn_index_sample.php
tim-wovn: with_ssi or without_ssi?

src/wovn_helper.php
```diff
@@ -3,11 +3,29 @@
...
+  }
+  return implode('/', $tmp_out);
+}
+
 function wovn_helper_detect_paths($base_dir, $path_of_url, $files) {
   $request_path = $base_dir . $path_of_url;
-  $local_path = realpath($request_path);
-  $inside_base_dir = $local_path && strpos($local_path, $base_dir) === 0;
-  $local_path = $inside_base_dir ? $local_path : $base_dir;
+  $local_path = remove_dots_from_path($request_path);
```
a-bougouin: Do you think `realpath` function can apply in this case, instead of making our own `remove_dots_from_path` function?

tim-wovn: Oh hmmmm...maybe the bug is in the behavior after realpath.  Let me try something.

wovn_index_sample.php
a-bougouin: By default please use the non-ssi inclusion.

> Is this a reasonable thing to do?

I am ok with this.

tim-wovn:
A follow-up on the previous comment.  That comment is true, except for HHVM.

We have a given behavior for some versions of PHP and a different behavior for others.  What should we do then?

src/wovn_helper.php
```diff
@@ -3,11 +3,32 @@
 require_once(__DIR__ . '/wovnio/wovnphp/SSI.php');
 use Wovnio\Wovnphp\SSI;
 
+function remove_dots_from_path($path) {
+  # Removes '/./ in paths and resolves '/../' in paths
+  # From https://tomnomnom.com/posts/realish-paths-without-realpath
+  # See also http://php.net/manual/en/function.realpath.php#84012
+  $path = str_replace('//', '/', $path);
```
hythof: In some case, `str_replace` do unexpected replacing.
```
$ php -r 'echo str_replace("//", "/", "///");'; echo
//
```

src/wovn_helper.php
hythof: I think that it's need prevent directory traversal.
For example, if user request path is `../../../../../../etc/passwd`, we don't permit return `/etc/passwd`.
But `realpath` does not consider such as security side and `remove_dots_from_path` is too.

tim-wovn:
Shoji said it can be deployed when we are ready.

tim-wovn:
@hythof Right now I have:
  $local_path = remove_dots_from_path($request_path);
  $local_path = realpath($local_path);

In this case, $request_path will first have '..' resolved.  In the case you provided, it will be 'etc/passwd'  Then realpath() will make this an absolute path based off of the path of the website, which I guess will not be '/'

1) Leave as is
2) Move realpath inside of remove_dots_from_path() to prevent other users from using it in the wrong way (although the function does what it claims to--it only removes dots from paths)
3) Do something else (I'm not sure how we should prevent directory traversal)
4) I've misunderstood

test/wovn_index_sample_workspace/WOVN.php/src/wovn_helper.php
```diff
@@ -3,11 +3,47 @@
...
+  return implode('/', $tmp_out);
+}
+
 function wovn_helper_detect_paths($base_dir, $path_of_url, $files) {
   $request_path = $base_dir . $path_of_url;
-  $local_path = realpath($request_path);
-  $inside_base_dir = $local_path && strpos($local_path, $base_dir) === 0;
-  $local_path = $inside_base_dir ? $local_path : $base_dir;
+  $local_path = remove_dots_from_path($request_path);
+  $local_path = realpath($local_path);
```
hythof: Now I think it's enough prevent directory traversal.
But I'm not sure 100%, so could you check if file of `$local_path` in directory of `$base_dir` after do `realpath`?
If not, we have to don't accept by security reason.

wovn_index_sample.php
```diff
@@ -14,11 +14,22 @@
...
+# (also see the SSI comment below)
 $included = wovn_helper_include_by_paths($paths);
 # $included = wovn_helper_include_by_paths_with_ssi($paths);
 
 # Set 404 status code if file not included
 if (!$included) {
   header($_SERVER["SERVER_PROTOCOL"] . " 404 Not Found");
-  echo "Page Not Found";
+
+  # Look for 404.html file in the root directory
```
hythof: 👍

tim-wovn:
@a-bougouin @hythof One final question for this PR: if a really bad URL is passed in, we still load in the base URL.  Should we perhaps instead load a 404?

E.g. should `http://stage-dev.tokyo-dome.co.jp/en///////tourists/aaaaaaa/b` load a 404 page or `http://stage-dev.tokyo-dome.co.jp`

For general bad urls the customer wants a 404 page loaded.  For crazy urls though, its not clear to me that they care or that other customers will care.

a-bougouin:
@tim-wovn I think malformed URL should be considered the same and redirect to 404.

tim-wovn:
@a-bougouin @hythof Ok, malformed URLs and well-formed, but bad URLs will both lead to a 404 page.

It seems like kindof a big change.  Our other customers will be ok with it?

If everything is ok, please let me know.  We can merge this and then deploy!

a-bougouin:
@hythof ?

wovn_index_sample.php
```diff
@@ -12,13 +12,33 @@
...
+$parsed_url = parse_url($_SERVER["REQUEST_URI"], PHP_URL_PATH);
+
+if ($parsed_url) {
+  $paths = wovn_helper_detect_paths(dirname(__FILE__), $parsed_url, $files);
+  
+  # SSI USER: please swap comments on the two lines below
+  # (also see the SSI comment in the 404 section below)
+  $included = wovn_helper_include_by_paths($paths);
+  # $included = wovn_helper_include_by_paths_with_ssi($paths);
+}
```
tim-wovn: The whitespace on lines 19 and 23 causes the SSI test to fail.  If those lines are flush with 'if' (they are untabbed), then the test passes.

hythof:
@tim-wovn I think that showing 404 page is better when URL is malformed
It's enough to tell to shoji about this Also customer can avoid this changing by they's .htaccess

PR #48, opened 2018-09-06
SSI resolved file import

a-bougouin:
Fix SSI inclusion to resolve path like we do for URI script filename.

### Comments
Because my fix includes `wovn_helper.php`, I encountered problem with `SSIWovnIndexSampleTest.php` because it runs its own WOVN.php environment although the other tests required `src` code. It lead into `require_once` not erroring out saying that same function was already declared from other file. To fix I separated the test into unit and integration tests which run in isolation.

Regular command `./vendor/bin/phpunit` will run unit tests and `./vendor/bin/phpunit --configure phpunit_integration.xml` will run integration tests (CircleCI is configured to run both).


src/wovnio/wovnphp/SSI.php
```diff
@@ -9,18 +11,31 @@ public static function readFile($includePath, $rootDir=null) {
     }
 
     public static function readFileRecursive($includePath, $rootDir, &$limit) {
-      $ssi_include_regexp = '/<!--#include virtual="(.+?)"\s*-->/';
+      $ssiIncludeRegexp = '/<!--#include virtual="(.+?)"\s*-->/';
```
suplo: as I see, we use both `$camelCase and `$this_var` conventions, should we use just only 1?

test/integration/SSIWovnIndexSampleTest.php
```diff
@@ -1,44 +1,46 @@
...
-    $sampleIndexFile = '../../wovn_index_sample.php';
+    $sampleIndexFile = '../../../wovn_index_sample.php';
     $inclusionCode = '\$included\ =\ wovn_helper_include_by_paths\(\$paths\)\;';
     $ssiInclusionCode = '\$included\ =\ wovn_helper_include_by_paths_with_ssi\(\$paths\)\;';
 
-    chdir(dirname(__FILE__) . '/wovn_index_sample_workspace');
+    mkdir($this->workspace);
+    chdir($this->workspace);
+
+    mkdir('WOVN.php');
```
suplo: could use `recursive` param for creating nested directories

src/wovnio/wovnphp/SSI.php
a-bougouin: Yes, tried to limit changes to minimum, so i used conventions that are used in each file. As I mentionned earlier I want to start cleaning up.

Thanks for comment.

PR #49, opened 2018-09-11
Cleanup Environment Test Fixture + Update HeadersTest

a-bougouin:
Cleanup test environment.

### Comments
I had time to update few tests currently, so hold fixtures are still there at `test/unit` directory.

This PR, add factory for creating `Store` and `Headers` from fixture files. Only `HeadersTest` is using it right now (60% of it).

It would like to merge it although everything is not cleaned up yet, so we can start using new helpers and fixtures for new tests.

### DIFF
--- a/test/unit/APITest.php
+++ b/test/unit/APITest.php
@@ -1,4 +1,6 @@
 <?php
+  require_once 'test/helpers/StoreAndHeadersFactory.php';
+
   require_once 'src/wovnio/wovnphp/API.php';
   require_once 'src/wovnio/wovnphp/Utils.php';
   require_once 'src/wovnio/wovnphp/Lang.php';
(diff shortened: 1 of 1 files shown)


test/helpers/StoreAndHeadersFactory.php
```diff
@@ -0,0 +1,47 @@
+<?php
+namespace Wovnio\Test\Helpers;
+
+use Wovnio\Wovnphp\Store;
+use Wovnio\Wovnphp\Headers;
+
+class StoreAndHeadersFactory {
+  public static function fromFixture($fixture = 'default', $settingsOverwrite = array(), $envOverwrite = array()) {
```
suplo: I think a Factory should be like

```
class StoreAndHeadersFactory {
  public function setSomething($option) { // could have many setSomething for each option like store, fixture, etc. Default values are set in constructor
    $this->$someOption = $option;
    return $this;
  }

  public function build() {
     // build store and header based on what are set
     ...

     return array($store, $headers)
  }
}
```

and being called like `new StoreAndHeadersFactory().setSomething($option).build()`

How do you think?

a-bougouin: This is not following best practice Factory Method. Or maybe you have a reference you can share? This approach also make it unreliable when used multiple time. I prefer the way it is now.

suplo: indeed, it's Builder pattern. I think this make it clear that when you want to use different options, you have to set it up explicitly, and have just 1 `build()`, instead of having multiple methods for each type of build (from fixture, custom store, etc.)

a-bougouin: I prefer using Factory Method here.

test/helpers/StoreAndHeadersFactory.php
```diff
@@ -0,0 +1,47 @@
...
+
+    return array($store, $headers);
+  }
+
+  private static function buildStoreOptions($options) {
+    $defaultOptions = array(
+      'default_lang' => 'en',
+      'supported_langs' => array('en'),
+      'url_pattern_name' => 'path',
+      'project_token' => 'KK9kZ'
```
toinoudivad: This is my User token :'( but I guess it's fine

a-bougouin: I took it from current tests, I'll use another one (hopefully there won't be a lot of tests to change).

PR #51, opened 2018-09-25
Use StoreAndHeadersFactory in API tests

a-bougouin:
Use StoreAndHeadersFactory in API tests

### DIFF
--- a/test/unit/APITest.php
+++ b/test/unit/APITest.php
@@ -43,7 +43,7 @@
       return $mockObject;
     }
 
-    public function mockApiResponse($expected_api_url, $expected_data, $response) {
+    private function mockApiResponse($expected_api_url, $expected_data, $response) {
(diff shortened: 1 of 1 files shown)


test/unit/APITest.php
```diff
@@ -43,17 +43,23 @@ private function getMockAndRegister($originalClassName, $methods) {
       return $mockObject;
     }
 
-    protected function setUp() {
-      RequestHandlerFactory::set_instance(NULL);
-    }
-
-    protected function tearDown() {
-      RequestHandlerFactory::set_instance(NULL);
+    public function mockApiResponse($expected_api_url, $expected_data, $response) {
```
mrasu: Can you make it private because it is not used at other files?

a-bougouin: Right, I will.

a-bougouin: Done, thanks for review.

PR #53, opened 2018-09-26
Use new fixtures for UtilsTest

a-bougouin:
Use new fixtures for UtilsTest


test/unit/UtilsTest.php
```diff
@@ -1,32 +1,29 @@
...
-
   public function testFunctionsExists() {
     $this->assertTrue(class_exists('Wovnio\Wovnphp\Utils'));
     $this->assertTrue(method_exists('Wovnio\Wovnphp\Utils', 'getStoreAndHeaders'));
     $this->assertFalse(method_exists('Wovnio\Wovnphp\Utils', 'dispatchRequest'));
   }
 
   public function testGetStoreAndHeaders() {
-    $env = $this->getEnv('_path');
+    $env = EnvFactory::fromFixture('default');
```
mrasu: Because "default" is default argument for `fromFixture`, not needed?

a-bougouin: Yes, I actually wanted to remove the default because `EnvFactory::fromFixture()` does not make too much sens to me. I'll remove default value.

a-bougouin: I removed default value, thanks for review

PR #58, opened 2018-10-16
Update Code with PSR2 Norm

a-bougouin:
Update Code with PSR2 Norm

### Comments
I got rid of as much WARNING as possible, but some still remain.
I would like to merge this PR as soon as possible since it removes a lot of errors.
Once every errors are fixed, I hope we can remove remaining warnings and set WARNINGS to be considered as errors (but it might not be possible).


a-bougouin:
I’ll make more readable PR.

PR #61, opened 2018-10-18
Apply Coding Standard to Headers

a-bougouin:
Apply coding standard to Headers

### Comments
I found that the rule for camelCased variables does not apply to methods, so I added a comment to explain which rule to use instead. I did not apply the rule yet because it requires fixing case on files that were already removed from `phpcs` blacklist (I'll activate this rule next if approved).

### DIFF
--- a/src/wovnio/wovnphp/Headers.php
+++ b/src/wovnio/wovnphp/Headers.php
@@ -200,7 +200,7 @@
      *
      * @return String The browser lang
      */
-    public function computerBrowserLang()
+    public function computeBrowserLang()
     {
       if ($this->browserLang === null) {
         // cookie lang


src/wovnio/wovnphp/Headers.php
```diff
@@ -229,20 +238,22 @@ public function browserLang() {
...
      * @return String The url of the redirections location
      */
-    public function redirectLocation($lang=null) {
-      if ($lang === null)
-        $lang = $this->browserLang();
-      if ($lang === $this->store->settings['default_lang'])
+    public function redirectLocation($lang = null)
+    {
+      if ($lang === null) {
+        $lang = $this->computeBrowserLang();
```
suplo: where is `computeBrowserLang` defined?

a-bougouin: Typo, I defined as `computerBrowserLang`, thanks.
It is sad that tests did not catch :'(

PR #69, opened 2018-10-30
Logger

a-bougouin:
### Purpose/goal of PR
Add a logger to homogenize log messages and allow to quiet WOVN.php logs.
Format is: `WOVN.php [MESSAGE_TYPE] MESSAGE`.

Inspired by PSR3: https://www.php-fig.org/psr/psr-3/.

### Comments
As current logs, `Logger` uses `error_log` to log messages. In the future, we might consider using `syslog` instead or allow user to set a separate log file for WOVN.php (`error_log('message', 3, 'path/to/custom/log/file')`).

### Release comments
Previous log format was `****** WOVN++ LOGGER :: MESSAGE ******`. If user parse logs, their parsing might no longer work. Do we need to do something about it? Is release change log enough? Or should I put back the logging to this format?

### DIFF
--- a/src/wovnio/wovnphp/API.php
+++ b/src/wovnio/wovnphp/API.php
@@ -4,8 +4,8 @@
   require_once DIRNAME(__FILE__) . '../../utils/request_handlers/RequestHandlerFactory.php';
 
   use \Wovnio\Wovnphp\Logger;
-  use Wovnio\Html\HtmlConverter;
-  use Wovnio\Utils\RequestHandlers\RequestHandlerFactory;
+  use \Wovnio\Html\HtmlConverter;
+  use \Wovnio\Utils\RequestHandlers\RequestHandlerFactory;
 
   if (!defined('WOVN_PHP_VERSION')) {
     define('WOVN_PHP_VERSION', '0.1.10');
--- a/src/wovnio/wovnphp/API.php
+++ b/src/wovnio/wovnphp/API.php
@@ -3,6 +3,7 @@
 
   require_once DIRNAME(__FILE__) . '../../utils/request_handlers/RequestHandlerFactory.php';
 
+  use \Wovnio\Wovnphp\Logger;
   use Wovnio\Html\HtmlConverter;
   use Wovnio\Utils\RequestHandlers\RequestHandlerFactory;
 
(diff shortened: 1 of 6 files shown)


a-bougouin:
> I think changing it should be ok as long as we explain it in the release notes.

@tim-wovn thanks

src/wovnio/wovnphp/API.php
```diff
@@ -3,6 +3,7 @@
 
   require_once DIRNAME(__FILE__) . '../../utils/request_handlers/RequestHandlerFactory.php';
 
+  use \Wovnio\Wovnphp\Logger;
   use Wovnio\Html\HtmlConverter;
```
suplo: should the next 2 lines be `use \Wovnio...`?

src/wovnio/wovnphp/Logger.php
```diff
@@ -0,0 +1,97 @@
...
+    if (self::$logger === null) {
+      self::$logger = new Logger();
+    }
+
+    return self::$logger;
+  }
+
+  public static function set($logger)
+  {
+    self::$logger = $logger;
```
suplo: should we throw error if `$logger` is not instance of Logger?

src/wovnio/wovnphp/API.php
a-bougouin: I think it is best to use `\` at begining of namespace here. I was afraid to change, but then I did it for an other file yesterday. I’ll change for more homogeneity :+1:

test/bootstrap.php
```diff
@@ -17,3 +18,6 @@
 require_once 'src/wovnio/modified_vendor/SimpleHtmlDomNode.php';
 
 require_once 'test/helpers/StoreAndHeadersFactory.php';
+
+// disable error logging
+\Wovnio\Wovnphp\Logger::set(new \Wovnio\Wovnphp\Logger(true));
```
suplo: could we just have a `quite` method to disable logger output?, then we don't really need `set` method

src/wovnio/wovnphp/Logger.php
a-bougouin: Yes we can. Currently it allows for duck typing, if our user want’s to use another PSR3 compatible logger for WOVN.php logs.

I don’t think adding an exception is a must have here and for the reason I just gave I prefer not.

Please feel free to argue if you disagree.

suplo: Ah, thanks for the explanation. Let's keep the code then

test/bootstrap.php
a-bougouin: I still want `set` method for the reason I mentionned in the above comment. 

I agree with `quiet` function, I’ll add it. What name should I use for “unquiet” function?

suplo: `unquite` is a good name imo 👯‍♂️

a-bougouin: I made the change. In the end I went for simple setter.

PR #70, opened 2018-10-31
Update Store coding standard

a-bougouin:
Update Store coding standard

### DIFF
--- a/src/wovnio/wovnphp/Store.php
+++ b/src/wovnio/wovnphp/Store.php
@@ -1,26 +1,32 @@
 <?php
   namespace Wovnio\Wovnphp;
+
   use Wovnio\Html\HtmlConverter;
 
   /**
    * The Store class contains the user settings
    */
-  class Store {
+  class Store
+  {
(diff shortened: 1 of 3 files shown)


src/wovnio/wovnphp/Store.php
```diff
@@ -1,26 +1,32 @@
...
-    public $config_loaded = false;
-    public static $config_dir;
-    public $return_value_on_error_handle = false;
+    // FIXME: could be private (unused outside this scope???)
+    public $configLoaded = false;
+    // FIXME: could remove (unused???)
+    public static $configDir;
+    // FIXME: could remove (unused???)
+    public $returnValueOnErrorHandle = false;
+    // FIXME: wovnphp remaining code???
```
suplo: why do you leave this FIXME? Does it take time to remove?

a-bougouin: Point is to change coding standard, not make other changes that could have side effect if i make mistake. So I annotate to fix later.

PR #84, opened 2019-06-05
Clarify that path htaccess rules should be put at begining of htaccess

a-bougouin:
### Purpose/goal of PR
Make doc more precise

### Comments
Since it is important that path htaccess rules are on top of htaccess to prevent interfering with `-d` and `-f` checks, I added it in documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ language information. WOVN.php supports three patterns.
 You need to change your server settings to strip the language codes off of the
 URL before it is processed by you scripts.
 
-For Apache users, you can add the following rules at the top of you `.htaccess`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -188,8 +188,8 @@ language information. WOVN.php supports three patterns.
 You need to change your server settings to strip the language codes off of the
 URL before it is processed by you scripts.
 
-For Apache users, you can add the following to your `.htaccess`. You will need
-to activate the `mod_rewrite` PHP module. Please follow the
+For Apache users, you can add the following rules at the top of you `.htaccess`.
```
torkelbe: rules -> rule
you -> your

a-bougouin: fixed, thanks

PR #88, opened 2019-06-14
Support lang param name feature

torkelbe:
Add support for `lang_param_name` feature.

The setting defaults to `wovn`, and it is declared in snippet settings always. (That is also why so many testing files had to change in this PR.)

Explanation from README:
> #### `lang_param_name`
> This parameter is only valid for when `url_pattern_name = query`.
> 
> It allows you to set the query parameter name for declaring the language of the page.
> The default value for this setting is `lang_param_name = wovn`, such that a page URL
> in translated language English has the form


src/wovnio/wovnphp/Store.php
```diff
@@ -125,7 +126,11 @@ public function updateSettings($vals)
 
         // getting the url pattern
         if (isset($vals['url_pattern_name']) && $vals['url_pattern_name'] === 'query') {
-            $vals['url_pattern_reg'] = '((\?.*&)|\?)wovn=(?P<lang>[^&]+)(&|$)';
+            if (isset($vals['lang_param_name'])) {
```
bstewart00: internally, can we make it so lang_param_name is always set to `wovn` by default? That just gets overridden if the user sets it? So we could remove this else branch.

torkelbe: Ugh, I agree this branch is ugly. The predicament comes from how the default values are set somewhere else and not accessed in this function, but we kind of need it for creating `url_pattern_reg`. The default value for the setting always set to `wovn`, but it is only merged after this function has returned.

I will not be doing any settings management refactoring here, but I can try to make branch a bit cleaner.

torkelbe: I effectively just moved the branch, and I think it got better. It still is ugly, but I want to work within the frame in which all the other settings are managed.

src/wovnio/wovnphp/Store.php
```diff
@@ -125,7 +126,13 @@ public function updateSettings($vals)
...
+                $lang_param_name = $vals['lang_param_name'];
+            } else {
+                $defaultSettings = $this->defaultSettings();
```
a-bougouin: `$this->defaultSettings()` are already loaded down in `updateSettings` so you should instanciate `$defaultSettings` at top scope for single call.

torkelbe: done

I was hesitant to initialize `$defaultSettings` at the top of the function because accessing it feels like a hack with the way settings are handled in Store.php at the moment. But you're right, it's just more dumb to initialize it twice. =/

a-bougouin: It does feel like a hack, yes.

PR #91, opened 2019-07-04
Support Custom Lang Aliases in `supported_langs`

a-bougouin:
### Purpose/goal of PR
Support users that use custom lang aliases in their list of supported languages.

### Comments
Some user mistake and input their custom language codes in the list of supported languages. This leads to invalid hreflangs being generated by WOVN.php. This change iterates through supported languages list and ensure the language codes are set to valid WOVN.php language codes, using user's custom lang aliases if necessary.

### DIFF
--- a/src/wovnio/wovnphp/Store.php
+++ b/src/wovnio/wovnphp/Store.php
@@ -42,23 +42,9 @@ class Store
      */
     public function __construct($userSettings)
     {
-        $defaultSettings = $this->defaultSettings();
+        $this->settings = $this->defaultSettings();
         if ($userSettings) {
-            $userSettings = $this->updateSettings($userSettings);
-        } else {
-            $userSettings = array();
-        }
-
-        $this->settings = array_merge($defaultSettings, $userSettings);
-
-        // Use default api_url property when user api_url property is empty.
-        if ($this->settings['api_url'] === '') {
-            $this->settings['api_url'] = $defaultSettings['api_url'];
-        }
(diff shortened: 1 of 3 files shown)


src/wovnio/wovnphp/Store.php
```diff
@@ -204,9 +220,13 @@ public function convertToCustomLangCode($lang_code)
         return $lang_code;
     }
 
-    public function convertToOriginalCode($lang_code)
+    public function convertToOriginalCode($lang_code, $customLangAliases = null)
```
bstewart00: Don't like this null default parameter, but I think the real problem is `updateSettings` taking in this $vals parameter. From what I can tell, $vals is expected to be the same object as this->settings. I would much rather store using this->settings always, if it is possible.

bstewart00: Maybe it could be done by doing `$this->settings = array_merge($defaultSettings, $userSettings);` first in the constructor, followed by `updateSettings`, then you wouldn't need $vals.

a-bougouin: I agree, I'll see what can be done.

a-bougouin: I fixed the initialization of Store so that my change can expect settings to be usable instead of adding to method prototype.

bstewart00:
Oh yeah, it looks like updateSettings doesn't need to be public as well.

PR #92, opened 2019-07-11
[FIX] cURL timeout error

Soichiro725:
Same as [[FIX] cURL timeout error](https://github.com/WOVNio/wordpress_plugin/pull/62)


Soichiro725:
Same as [[FIX] cURL timeout error](https://github.com/WOVNio/wordpress_plugin/pull/62)

PR #93, opened 2019-07-31
Add Tagalog Support

a-bougouin:
### Purpose/goal of PR
Add Tagalog Support


Soichiro725:
I found a lang code in `STATIC_DOC.txt`.
It is better to change this file also.

a-bougouin:
@Soichiro725 thanks. This file is actually useless since WOVN.php README was renewed, so I deleted it.

PR #96, opened 2019-10-01
Prevent Interception Errors Due to CWD Changed

a-bougouin:
### Purpose/goal of PR
Prevent interception errors due to CWD changed

### Comments
When WOVN.php intercepts through wovn_index.php, CWD is the root because wovn_index.php is at the root. However, the file intercepted by WOVN.php is not necessary at the root and relative imports become broken. This PR set the CWD to the directory of the file intercepted.


Soichiro725:
Can you give me a sample which cause problem?

PR #100, opened 2019-11-08
Add tests for addValue and revert

Soichiro725:
Same fix with the following PR.
https://github.com/WOVNio/wovnrb/pull/152

WOVN.php already have fix, but there are no tests.
So, I just add tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
   "name": "WOVNio/WOVN.php",
-  "type": "Library",
+  "type": "library",
   "description": "PHP server library for WOVNio",
   "homepage": "https://wovn.io",
   "require": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "WOVNio/WOVN.php",
-  "type": "Library",
+  "type": "library",
```
Soichiro725: The docker image of circleci/php has changed recently.
Maybe composer was upgrated and it starts to cause the following error.
So, I changed to use lowercase.
```
  [Composer\Json\JsonValidationException]                     
  "./composer.json" does not match the expected JSON schema:  
   - type : Does not match the regex pattern ^[a-z0-9-]+$  
```

bstewart00:
Interesting that wovn.php and wovn.cs already have the `addValue` function. Maybe it was only wovn.rb that was missing?

PR #103, opened 2020-01-09
[MISC]WOVN.php improvement

Soichiro725:
https://github.com/WOVNio/equalizer/issues/25125

- Delete `WOVNPHP_DEFAULT_INDEX_FILE`
- Add docker environment with Wordpress
- Add sample to not allow access `wovn.ini` to `htaccess_sample` 
- Add `.htaccess` to not allow access WOVN.php directly
- Reverse CircleCI tests order of PHP versions to find error version easily

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,1 @@
+Deny from all
--- a/.htaccess
+++ b/.htaccess
@@ -1,1 +1,2 @@
+# Block direct access to WOVN.php directory
 Deny from all
--- a/htaccess_sample
+++ b/htaccess_sample
@@ -1,3 +1,8 @@
+# Don't allow access to wovn.ini direcly
+<Files ~ "^(wovn\.ini)$">
+  deny from all
+</Files>
+
 <IfModule mod_rewrite.c>
   RewriteEngine On
   # For path pattern, remove language code
--- a/htaccess_sample
+++ b/htaccess_sample
@@ -1,4 +1,4 @@
-# Don't allow access to wovn.ini direcly
+# Don't allow access to wovn.ini directly
 <Files ~ "^(wovn\.ini)$">
   deny from all
 </Files>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,20 +16,20 @@ jobs:
           key: vendor-{{ checksum "composer.lock" }}
           paths: vendor
       - run: ./vendor/bin/phpcs
-      - run: scripts/run_phpunit_tests.sh vectorface/php5.3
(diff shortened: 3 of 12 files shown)


htaccess_sample
```diff
@@ -1,3 +1,8 @@
+# Don't allow access to wovn.ini direcly
```
bstewart00: typo `direcly` -> `directly`

bstewart00:
This doesn't actually change anything right? The existing code had the `WOVNPHP_DEFAULT_INDEX_FILE` check. (Other than making it able to be a single string)

I wonder if this should be a wovn.ini setting instead?

Soichiro725:
> This doesn't actually change anything right? The existing code had the `WOVNPHP_DEFAULT_INDEX_FILE` check. (Other than making it able to be a single string)
> 
> I wonder if this should be a wovn.ini setting instead?

I just allowed to set `WOVNPHP_DEFAULT_INDEX_FILE` with array.
But PHP 5.3 doesn't allow to set array with `define`.

I thought `wovn.ini` is good to set this config.
But there are some problems.
- Now `src/wovn_helper.php` doesn't have a dependency of Store, so I don't want to add dependency.
- It's better to not delete `WOVNPHP_DEFAULT_INDEX_FILE` to keep compatibility.

So, I'm thinking to add file name array option to `wovn_helper_detect_paths`.

.htaccess
```diff
@@ -0,0 +1 @@
+Deny from all
```
a-bougouin: This new file is mistake?

Soichiro725: I wanted to block direct access to WOVN.php directory.

a-bougouin: Ah, I see

bstewart00:
From the history it looked to me that the define variable was only added for test purposes, so I would question if there is really a backwards compatibility issue

Soichiro725:
> From the history it looked to me that the define variable was only added for test purposes, so I would question if there is really a backwards compatibility issue

Yes. You are right.
Almost all users don't use WOVNPHP_DEFAULT_INDEX_FILE.
Even if user uses it, problem will happen when library is updated, but I think that updating is few times.
So I actually want to delete WOVNPHP_DEFAULT_INDEX_FILE.
Then, I will delete WOVNPHP_DEFAULT_INDEX_FILE.

Soichiro725:
Finally, I just deleted `WOVNPHP_DEFAULT_INDEX_FILE`.
I will ask user to customize `wovn_helper.php` directly.
It is enough and simple.

PR #104, opened 2020-01-31
Remove unnecessary explain for wovn users

iwai-wovn:

Remove because it is relevant to the wovn user


bstewart00:
It's kind of useful for developers though 😢

PR #105, opened 2020-04-15
[HOTFIX] Adjust doc file regexp

iwai-wovn:
### Purpose/goal of PR

Document files (e.g. doc/xls) are not eligible for translation.

### DIFF
--- a/src/wovnio/wovnphp/Utils.php
+++ b/src/wovnio/wovnphp/Utils.php
@@ -171,8 +171,9 @@ class Utils
         $uri_path = preg_replace("/^(https?:\/\/)?/", "", $uri);
         // strip host
         $uri_path = preg_replace("/^[^\/]*/", "", $uri_path);
-        // strip query
-        $uri_path = preg_replace("/\?(.)*$/", "", $uri_path);
+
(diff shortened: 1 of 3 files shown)


src/wovnio/wovnphp/Utils.php
```diff
@@ -169,4 +175,9 @@ private static function getPath($uri)
         $uri_path = preg_replace("/\?(.)*$/", "", $uri_path);
```
bstewart00, line 175: Could replace this with removeQueryAndHash. I think it makes sense to remove hash as well.

Soichiro725:
@bstewart00 
Do you think this PR is needed? As you suggested, this PR will be replaced by `Content-Type` way.

bstewart00:
It's a tiny PR so I think it's fine to do this first even if we replace it later.

Soichiro725:
> It's a tiny PR so I think it's fine to do this first even if we replace it later.

OK

src/wovnio/wovnphp/Utils.php
iwai-wovn: Thx! change this line.

PR #108, opened 2020-05-18
Make it easy to create dev env

Soichiro725:
I changed the followings.
- Add makefile
- Move yml files into docker directory
- Make widget URL changeable via WOVN.ini


makefile
```diff
@@ -0,0 +1,13 @@
+DOCKER_COMPOSE_YML = docker/apache.yml
+
+build:
```
a-bougouin, line 3: I would add `.PHONY` declaration for `build`, `stop`, `start` and `clean`, because those recipes aren't existing files:
```suggestion
.PHONY: build stop start clean

build:
```

src/wovnio/html/HtmlConverter.php
```diff
@@ -196,13 +196,11 @@ private function buildSnippetCode($adds_backend_error_mark)
...
+        $widget_url = $this->store->settings['widget_url'];
+        $fallback_mark = $adds_backend_error_mark ? 'data-wovnio-type="fallback_snippet"' : '';
         $data_wovnio = htmlentities("key=$token&backend=true&currentLang=$current_lang&defaultLang=$default_lang&urlPattern=$url_pattern&langCodeAliases=$lang_code_aliases_json&langParamName=$lang_param_name&version=WOVN.php");
 
-        if ($adds_backend_error_mark) {
-            return "<script src=\"//j.wovn.io/1\" data-wovnio=\"$data_wovnio\" data-wovnio-type=\"fallback_snippet\" async></script>";
-        } else {
-            return "<script src=\"//j.wovn.io/1\" data-wovnio=\"$data_wovnio\" async></script>";
-        }
+        return "<script src=\"$widget_url\" data-wovnio=\"$data_wovnio\" $fallback_mark async></script>";
```
bstewart00, line 203: i'm curious where the extra space is coming from in the tests. It doesn't look like there should be a space here

Soichiro725: Extra space comes from both sides of `$fallback_mark` when `$fallback_mark` is empty.
I didn't want to write similar snippet twice.
Do you think I should change as before?

bstewart00: Getting rid of the extra space would be nice but not a big deal.
Maybe you could set `$fallback_mark = $adds_backend_error_mark ? ' data-wovnio-type="fallback_snippet"' : '';` to have a space on the front.
and do this?
`return "<script src=\"$widget_url\" data-wovnio=\"$data_wovnio\"$fallback_mark async></script>";`

Soichiro725: OK. I will do that.

PR #109, opened 2020-05-21
Add SitePrefixPath feature

Soichiro725:
Add sitePrefixPath feature.

- only valid for `path` pattern
- only translate pages that match the prefix path
- the path language code will be added after the prefix path
- don't insert snippet for URL which doesn't match the prefix path

e.g.
  swap to `en` with site_prefix_path = dir1/dir2 
  - http://example.com -> not insert snippet and not swap
  - http://example.com/dir1/dir2 -> http://example.com/dir1/dir2/en


README.en.md
```diff
@@ -419,6 +419,18 @@ Therefore, we do not add any WOVN script tags.
 check_amp = 1
 ```
 
+#### `site_prefix_path`
+This parameter tells WOVN.php to process under the path which is sepecified.
```
bstewart00, line 423:
```
This parameter tells WOVN.php to only process requests under the specified path. When translating URLs, the language identifier will be inserted after this prefix path.
This parameter is valid only when `url_pattern_name` is `path`
```
Just cleaned it up a bit

src/wovn_interceptor.php
```diff
@@ -22,8 +22,11 @@
 $_ENV['WOVN_TARGET_LANG'] = $headers->lang();
 $headers->requestOut();
 
-if (!Utils::isFilePathURI($headers->getDocumentURI(), $store) &&
-    !Utils::isIgnoredPath($headers->getDocumentURI(), $store)) {
+$uri = $headers->getDocumentURI();
+if (!Utils::isFilePathURI($uri, $store) &&
```
bstewart00, line 26: I would be OK with moving `shouldIgnoreBySitePrefixPath` inside `isIgnoredPath`. A path is ignored if it does not match the prefix path, after all.

src/wovnio/wovnphp/Store.php
```diff
@@ -117,14 +118,19 @@ private function updateSettings($updatedOptions)
...
         // getting the url pattern
         if ($this->settings['url_pattern_name'] === 'query') {
             $this->settings['url_pattern_reg'] = '((\?.*&)|\?)' . $this->settings['lang_param_name'] . '=(?P<lang>[^&]+)(&|$)';
         } elseif ($this->settings['url_pattern_name'] === 'subdomain') {
             $this->settings['url_pattern_reg'] = '^(?P<lang>[^.]+)\.';
         } else {
             $this->settings['url_pattern_name'] = 'path';
-            $this->settings['url_pattern_reg'] = '\/(?P<lang>[^\/.]+)(\/|\?|$)';
+            $prefix = empty($this->settings['site_prefix_path']) ? '' : str_replace('/', '\/', '/' . $this->settings['site_prefix_path']);
+            $this->settings['url_pattern_reg'] = $prefix . '\/(?P<lang>[^\/.]+)(\/|\?|$)';
```
bstewart00, line 133: I don't really like asserting on url_pattern_reg's regex as this ought to be an implementation detail. Store tests should be checking that we trim the trailing slash etc, but we should test the regex at a higher level - if a URL is processed or not with site prefix path.

src/wovnio/wovnphp/Utils.php
```diff
@@ -105,6 +105,18 @@ public static function isIgnoredPath($uri, $store)
         );
     }
 
+    public static function shouldIgnoreBySitePrefixPath($uri, $store)
+    {
+        if ($store &&
```
bstewart00, line 110: This seems overly defensive. When will store be null? If there is a case, I would prefer we do an early return in wovn_interceptor and assume everywhere else that store is non-null. Headers.requestOut already assumes it is non-null.

Can we also assume that Store will initialize `site_prefix_path` and `url_pattern_name` keys?

src/wovnio/wovnphp/Url.php
```diff
@@ -38,16 +38,21 @@ public static function addLangCode($uri, $store, $lang, $headers)
 
         $new_uri = $uri;
         $pattern = $store->settings['url_pattern_name'];
+        $site_prefix_path = $store->settings['site_prefix_path'];
         $lang_code = $store->convertToCustomLangCode($lang);
         $lang_param_name = $store->settings['lang_param_name'];
 
-        $no_lang_uri = self::removeLangCode($uri, $pattern, $lang_code, $lang_param_name);
-        $no_lang_host = self::removeLangCode($headers->host, $pattern, $lang_code, $lang_param_name);
+        if ($pattern === 'path' && !empty($site_prefix_path) && !self::matchSitePrefixPath($uri, $site_prefix_path)) {
```
bstewart00, line 45: This looks like a duplicate of `shouldIgnoreBySitePrefixPath`

src/wovnio/wovnphp/Utils.php
bstewart00: By the way, it seems strange to me that we do `headers->requestOut` before we check if the request is ignored.

src/wovnio/wovnphp/Url.php
```diff
@@ -174,8 +183,40 @@ public static function removeLangCode($uri, $pattern, $lang_code, $lang_param_na
...
+    }
+
+    private static function addPathLangCode($no_lang_url, $lang, $site_prefix_path = '')
+    {
+        if (empty($lang)) {
+            return $no_lang_url;
+        }
+        $prefix = $site_prefix_path ? '/' . $site_prefix_path : '';
+        return preg_replace(
+            '@' .
```
bstewart00, line 212: any way to re-use this regex with the one in `matchSitePrefixPath`?

README.en.md
Soichiro725: Done

src/wovn_interceptor.php
Soichiro725: Done

src/wovnio/wovnphp/Url.php
Soichiro725: `shouldIgnoreBySitePrefixPath` using `matchSitePrefixPath` inside.

src/wovnio/wovnphp/Utils.php
Soichiro725: > Can we also assume that Store will initialize site_prefix_path and url_pattern_name keys?

We initialize keys. So, We can assume there are `site_prefix_path` and `url_pattern_name` keys.

> By the way, it seems strange to me that we do headers->requestOut before we check if the request is ignored.

Ah, yes. This code means that WOVN will remove lang code even if the REQUEST_URI is ignored.

src/wovnio/wovnphp/Url.php
bstewart00: Not sure what you mean. This is exactly the same as what `shouldIgnoreBySitePrefixPath` is doing. It would be better if we made one of these functions private.

bstewart00: Actually, shouldn't this be calling `isIgnoredPath`?

src/wovnio/wovnphp/Store.php
Soichiro725: At StoreTest.php, I just added test of checking `url_pattern_reg` param.
I added this regex's tests at UrlTest.php and UtilsTest.php .
These tests run Store.php internally.

bstewart00: That's what I mean. UtilsTest is already testing the outcomes of this regex. It isn't really useful to assert that we are setting a specific regex.

src/wovnio/wovnphp/Url.php
Soichiro725: Done

src/wovnio/wovnphp/Store.php
Soichiro725: I changed tests to not check `url_pattern_reg` param directly.

src/wovnio/wovnphp/Url.php
Soichiro725: I agree with using `isIgnoredPath` is better to not modify ignored URL.
But this method is used everywhere.
I don't want to fix it this time because of fear of bad effects.

bstewart00: `isIgnoredPath` is only used once in wovn_interceptor.php. I don't see a problem with using it.

bstewart00: I think it is actually a bug if we don't use it. If the server redirects to `/ignored_path/`, we should not add lang code to it.

Soichiro725: Yes. That's right.
I will replace it.

src/wovnio/wovnphp/Url.php
```diff
@@ -174,8 +183,45 @@ public static function removeLangCode($uri, $pattern, $lang_code, $lang_param_na
...
                 break;
         }
     }
+
+    public static function shouldIgnoreBySitePrefixPath($uri, $settings)
+    {
+        if ($settings['site_prefix_path'] &&
+            $settings['url_pattern_name'] &&
+            $settings['url_pattern_name'] === 'path'
+        ) {
```
bstewart00, line 197: You said the store initializes default settings, so we should be able to remove 
```
$settings['site_prefix_path'] &&
            $settings['url_pattern_name'] &&
```
right?

src/wovnio/wovnphp/Url.php
```diff
@@ -174,8 +183,45 @@ public static function removeLangCode($uri, $pattern, $lang_code, $lang_param_na
...
             default:
                 // limit to one replacement
-                return preg_replace('/\/' . $lang_code . '(\/|$)/i', '/', $uri, 1);
+                $prefix = empty($site_prefix_path) ? '' : '/' . $site_prefix_path;
+                return preg_replace("@$prefix/$lang_code(/|$)@i", "$prefix/", $uri, 1);
                 break;
         }
     }
+
+    public static function shouldIgnoreBySitePrefixPath($uri, $settings)
```
bstewart00, line 192: I think it would be better if you move this to Utils.php as a private function used by isIgnoredPath + move the tests. We should only be calling `isIgnoredPath`

src/wovnio/wovnphp/Url.php
Soichiro725: `$settings['site_prefix_path']` could be null after initlization.
I will remove `$settings['url_pattern_name']`.

src/wovnio/wovnphp/Url.php
Soichiro725: `shouldIgnoreBySitePrefixPath` uses `generateUrlRegex` method which is defined at Url.php.
I didn't want to make `generateUrlRegex` public.
And I think `generateUrlRegex` should be in `Url.php`.

bstewart00: ok it's not too important as long as we only use isIgnoredPath. Could move either everything to Url.php or to Utils.php

src/wovnio/wovnphp/Url.php
Soichiro725: Done

src/wovnio/wovnphp/Utils.php
Soichiro725: > By the way, it seems strange to me that we do headers->requestOut before we check if the request is ignored.

I will split PR with more tests to change this.

src/wovnio/wovnphp/Url.php
Soichiro725: Done

src/version.php
```diff
@@ -3,5 +3,6 @@
     define('WOVN_PHP_NAME', 'WOVN.php');
 }
 if (!defined('WOVN_PHP_VERSION')) {
-    define('WOVN_PHP_VERSION', '0.1.23');
+    $version = isset($_ENV['WOVN_ENV']) && $_ENV['WOVN_ENV'] === 'development' ? 'VERSION' : '0.1.24';
```
bstewart00, line 6: Just noticed that we should start increment minor version number (0.2.0) for these kinds of features.

Soichiro725: sorry, I didn't notice this.

bstewart00: no problem, can do it next time

Soichiro725: How about increment to `1.0.0`.
Because this is not beta version.

bstewart00: Yes, definitely a good idea

PR #111, opened 2020-06-04
Add environment variables of ENV['WOVN_CONFIG']

iwai-wovn:
Allows a single WOVN.PHP library to dynamically switch between multiple settings.

Users can set `ENV['WOVN_CONFIG']` before loading `wovn_interceptor.php`.
```
$_ENV['WOVN_CONFIG'] = ... your config path ...;

require_once('/path/to/WOVN.php/src/wovn_interceptor.php');
```
Or, users can use `.htaccess` to set it with `SetEnv`. in the following
```
SetEnv WOVN_CONFIG /path/to/wovn.ini
```

### Comments
If ENV['WOVN_CONFIG'] is not defined, it falls back to the default `wovn.ini`.


iwai-wovn:
@bstewart00 
WOVN.PHP seems to embed wovn's script tag using the default value, even if `wovn.ini` doesn't exist, which is not the behavior I'm expecting.

If the user doesn't even have `wovn.ini`, I think it's better to do nothing, what do you think?
```
<script src="//j.wovn.io/1" data-wovnio="key=&amp;backend=true&amp;currentLang=en&amp;defaultLang=en&amp;urlPattern=query&amp;langCodeAliases=[]&amp;langParamName=wovn&amp;version=WOVN.php" data-wovnio-type="fallback_snippet" async=""></script>
```

bstewart00:
WOVN.php should abort if the settings file isn't available. This is what other backends do

README.en.md
```diff
@@ -419,7 +419,41 @@ Therefore, we do not add any WOVN script tags.
...
 
-## 4. Bug Report
+## 4. Environment Variable
+
+### `WOVN_TARGET_LANG`
+
+This environment variable sets the language code of the translation target as 
+retrieved from the HTTP request.
+The user can get the target language code from this environment variable and 
+arbitrarily change the behavior of your program.
```
a-bougouin, line 429:
```suggestion
The user can get the target language code from this environment variable and 
arbitrarily change the behavior of their program.
```

README.en.md
```diff
@@ -419,7 +419,41 @@ Therefore, we do not add any WOVN script tags.
...
+For example, you can make changes as follows
+
+Users can set `$_ENV['WOVN_CONFIG']` before loading `wovn_interceptor.php`.
+```
+$_ENV['WOVN_CONFIG'] = ... your config path ...;
+
+require_once('/path/to/WOVN.php/src/wovn_interceptor.php');
+```
+
+Or, users can use `.htaccess` to set it with `SetEnv`. in the following
```
a-bougouin, line 451:
```suggestion
Or, users can use `.htaccess` to set it with `SetEnv`:
```
Or
```suggestion
Or, users can use `.htaccess` to set as follows.
```

test/unit/StoreTest.php
```diff
@@ -37,6 +37,50 @@ public function testStoreExists()
...
+    {
+        $file_config = dirname(__FILE__) . '/test_config.ini';
+        if (file_exists($file_config)) {
+            unlink($file_config);
+        }
+        $data = 'project_token = ""' . "\n" .
+              'default_lang = en' . "\n" .
+              'supported_langs[] = en' . "\n" .
+              'supported_langs[] = ja' . "\n";
+        file_put_contents($file_config, $data);
```
Soichiro725, line 60: I prefer to use `implode` method.
```
implode("\n", array(
  'project_token = ""',
  ...
));
```

src/wovnio/wovnphp/Store.php
```diff
@@ -48,6 +43,20 @@ public function __construct($userSettings)
...
+     *  Configuration validate
+     *
+     *  @return boolean
+     */
+    public function isValid()
+    {
+        return (
+            $this->settings['project_token']
+            && $this->settings['default_lang']
+            && $this->settings['supported_langs']
```
Soichiro725, line 56: I think `url_pattern` also required.

iwai-wovn: The default value of `url_pattern_name` is `query`, so it is not necessary to check it.

bstewart00: I agree, we usually expect customers to specify url_pattern themselves

Soichiro725:
Can you also add description into `README.ja.md` ?

src/wovnio/wovnphp/Store.php
iwai-wovn: Oh, but you should check `url_pattern_name`, it's required.

src/wovnio/wovnphp/Utils.php
```diff
@@ -11,7 +11,7 @@ class Utils
     // will return the store and headers objects
     public static function getStoreAndHeaders(&$env)
     {
-        $file = DIRNAME(__FILE__) . '/../../../../wovn.ini';
+        $file = isset($_ENV['WOVN_CONFIG']) ? $_ENV['WOVN_CONFIG'] : DIRNAME(__FILE__) . '/../../../../wovn.ini';
```
Soichiro725, line 14: This means that user should set absolute path to $_ENV['WOVN_CONFIG'], right?
I want to add description about absolute path to README.

iwai-wovn:
> Can you also add description into `README.ja.md` ?

Yes, I was going to add it after I finished reviewing the English version.

src/wovnio/wovnphp/Utils.php
iwai-wovn: Added https://github.com/WOVNio/WOVN.php/pull/111/files#diff-4b29e98dc086ca7ee0e534c9eb921739R456

PR #113, opened 2020-06-10
Add Khmer language support

aw-leigh:
https://wovnio.atlassian.net/browse/ST-381
### Comments
Add Khmer language support, increment version


src/version.php
```diff
@@ -3,5 +3,6 @@
     define('WOVN_PHP_NAME', 'WOVN.php');
 }
 if (!defined('WOVN_PHP_VERSION')) {
-    define('WOVN_PHP_VERSION', '0.1.23');
+    $version = isset($_ENV['WOVN_ENV']) && $_ENV['WOVN_ENV'] === 'development' ? 'VERSION' : '0.1.25';
```
Soichiro725, line 6: Let's upgrade to `1.0.0`.
This library is not beta version now.

bstewart00: i'm thinking just do this version as-is. Then a separate PR to make version 1.0.0

PR #114, opened 2020-06-12
Refactoring integration tests

iwai-wovn:
- Replacing Integration tests with phpunit from shell scripts
- Changed to parallel test execution to support CircleCI's Workflows
- ~~Merged `scripts/run_server_tests.sh` into `scripts/run_phpunit_tests.sh`.~~
- Added generate a coverage report.
### Comments
- Parallelization has made the test run time a little faster.
- Coverage report is generated by PHP version 7 and above
- You can see the coverage report in the `ARTIFACTS` tab of CircleCI
- Using CircleCI's `Workflows`, the `store_test_results` does not seem to work


scripts/run_phpunit_tests.sh
```diff
@@ -4,14 +4,82 @@ set -eux
...
+    docker_name="php:5.3-apache"
+    # FIX: Unable to find image 'php:5.3-apache' locally and Error response from daemon: unexpected EOF
+    for n in $(seq 1 10); do docker pull $docker_name && break; done
+elif [ "${docker_name}" == "vectorface/hhvm" ]; then
+    # TODO: Not supported yet.
+    echo Not supported integration test for hhvm
+    exit 0
+fi
+
+# Run Apache for integration test
```
Soichiro725, line 52: I prefer to split integration test from `scripts/run_phpunit_tests.sh`.

test/integration/WovnIndexSampleApacheTest.php
```diff
@@ -0,0 +1,225 @@
...
+    {
+        $this->writeFile('wovn.ini', $contents);
+    }
+
+    private function setHtaccess($contents)
+    {
+        $this->writeFile('.htaccess', $contents);
+    }
+
+    private function writeFile($file, $contents)
```
Soichiro725, line 191: How about define writeFile at Utils.php.

test/integration/WovnIndexSampleApacheTest.php
```diff
@@ -0,0 +1,225 @@
...
+
+        mkdir($this->docRoot . '/v0');
+        copy($this->sourceDir . '/test/fixtures/integration/v0/translation', $this->docRoot . '/v0/translation');
+
+        chdir($this->docRoot);
+    }
+
+    protected function tearDown()
+    {
+        Utils::cleanUpDirectory($this->docRoot);
```
Soichiro725, line 28: I prefer cleaning up at the first of setUp.
Because If some test fails, after tests affect the failed tests.

test/integration/WovnIndexSampleApacheTest.php
```diff
@@ -0,0 +1,225 @@
...
+
+    private function writeFile($file, $contents)
+    {
+        $filePath = $this->docRoot . '/' . $file;
+
+        $content = is_array($contents) ? implode("\n", $contents) : $contents;
+        file_put_contents($filePath, $content);
+    }
+
+    private function getWovnIni($options = array())
```
Soichiro725, line 199: I think this method doesn't work for `custom lang code`.

scripts/run_phpunit_tests.sh
iwai-wovn: @Soichiro725 
In the case of a separate Integration Test, the creation of a dummy container and a code copy process must be performed again.
Initially they were isolated, but these did not help improve the speed of the test.
So I thought one script would be enough, what do you think?

test/integration/WovnIndexSampleApacheTest.php
iwai-wovn: How can I manage the path of the document root?
I have a feeling that the code would be redundant, as shown below.
```
Utils::writeFile($this->docRoot . ‘/‘ . ‘wovn.ini’, $content);
Utils::writeFile($this->docRoot . ‘/‘ . ’static.html’, $content2);
```

test/integration/WovnIndexSampleApacheTest.php
```diff
@@ -0,0 +1,225 @@
...
+        $contents = array();
+        foreach ($options as $name => $value) {
+            if (is_array($value)) {
+                foreach ($value as $v) {
+                    $contents[] = "${name}[] = $v";
+                }
+            } else {
+                $contents[] = "$name = $value";
+            }
+        }
```
iwai-wovn, line 221: @Soichiro725 
> I think this method doesn't work for custom lang code.

So custom lang code is a problem because it's structured like this, right?
`['custom_lang_aliases'] = array('ja' => ' Japanese');`
I am going to change the code as follows, what do you think?
```suggestion
        foreach ($options as $name => $value) {
            if (is_array($value)) {
                foreach ($value as $key => $v) {
                    if (is_string($key)) {
                        $contents[] = "${name}[$key] = $v";
                    } else {
                        $contents[] = "${name}[] = $v";
                    }
                }
            } else {
                $contents[] = "$name = $value";
            }
        }
```

scripts/run_phpunit_tests.sh
Soichiro725: I feel `scripts/run_phpunit_tests.sh` is too complex.
WOVN.php test is not so slow now.
And you can run integration tests in parallel by defining another workflows of circleCI.

test/integration/WovnIndexSampleApacheTest.php
Soichiro725: I prefer `Utils::writeFile($this->docRoot . ‘/‘ . ‘wovn.ini’, $content);` .
I can easily understand where is the file path.

test/integration/WovnIndexSampleApacheTest.php
Soichiro725: I think converting from setting to ini file cause complexity.
How about remove `getWovnIni` method and create wovn.ini via `setWovnIni`.
```
private function setHtaccess($options)
{
  $defaultOptions = array(
    'project_token => "Tek3n"',
    'api_url = "http://localhost/v0/"'
  );
  $merged_options = array_merge($defaultOptions, $options);
  $wovn_ini = implode("\n", $merged_options);
  $this->writeFile('wovn.ini', $wovn_ini);
}
```

Soichiro725: Is this solved?

.gitignore
```diff
@@ -6,6 +6,7 @@
 vendor/
 composer.phar
 test/integration/test/integration/wovn_index_sample_workspace
+.phpunit/
```
Soichiro725, line 9: Is this needed?

iwai-wovn: Coverage report is now outputted to .phpunit. If I create coverage report locally, we should not commit it by mistake.

Soichiro725: Ah, I think this commit is OK.
It is useful to develop locally.

PR #116, opened 2020-06-17
Dont replace script tag with ldjson

Soichiro725:
This is related to `schema translation` feature.
Previously, WOVN.php replace script tags and doesn't pass script tags to html-swapper.
This is because using cache as much as possible.
Some script tags is changed every time accessed.

This PR will avoid to replace by WOVN.php only for script tags which have `type="application/ld+json"`.

I also refactor tests in APITest.php to check the way to modify html easily.


src/version.php
```diff
@@ -3,6 +3,6 @@
     define('WOVN_PHP_NAME', 'WOVN.php');
 }
 if (!defined('WOVN_PHP_VERSION')) {
```
bstewart00, line 5: Minor bump (0.2.0). We definitely need to move wovn.php to 1.0.0 soon as well though.

Soichiro725: I changed 0.2.0 .
I will bump up to 1.0.0 soon after relase this PR.

Soichiro725:
> I find the tests a bit hard to read. Maybe something we should think about later is splitting up API and generating the converted HTML. So we could just unit test the different HTML scenarios without caring about mocking the API.

That's good.

> Also noticed we aren't using a mock library like https://github.com/mockery/mockery

Once, I tried to use Mockery. But there was a PHP version issue.
Mockery doesn't support version which we support.

PR #120, opened 2020-08-20
Feature/add diagnostic tools

zeyuwu:
This PR adds a diagnostic tools that made fast debugging on live systems easy.

### Comments
This PR added a diagnostics tool to WOVN.php that when ran, collects necessary information for debugging and displays the result. This diagnostic tool is not enabled by default (protected by user-selected username and password pairs.).

Added diagnostic tools


src/diagnostics.php
```diff
@@ -0,0 +1,217 @@
...
+                        if (file_put_contents($filepath, $buffer, LOCK_EX)) {
+                            return '<span style="color: #40b87c;">Diagnostic result has been saved successfully!</span>';
+                        } else {
+                            return '<span style="color: #ff5f5f;">Diagnostic result could not be saved, please check the file permissions (chmod a+w ' . $filepath . ')!</span>';
+                        }
+                    }
+                });
+                echo getAll(getAuthUser(), getAuthPassword());
+            } else {
+                echo '<span style="color: #ff5f5f;">The file ' . $filename . ' does not exist! <a href="/wovnphp/diagnostics.php">Try again</a></span>';
```
Soichiro725, line 209: `href="/wovnphp/diagnostics.php"` -> `href="/WOVN.php/diagnostics.php"`

README.en.md
```diff
@@ -494,3 +500,43 @@ as possible, we would need to know information like the followings.
...
+3. Create an empty `wovnphp_diagnostics.html` file inside the WOVN.php folder.
+
+4. Access the diagnostic tool at `yourwebsite.com/WOVN.php/diagnostics.php`. You will need to login using the password created in step 2.
+
+5. Follow the on screen instructions to get the diagnostic report, which will be written to the `wovnphp_diagnostics.html`.
+
+Please only enable the diagnostics tool when it is necessary to do so.
+
+#### If you are upgrading
+If you are upgrading from an older version of WOVN.php, please manually add the following lines to your `.htaccess` files:
```
a-bougouin, line 523: If this text stays for a while, it is better to state what "older version" means.
```suggestion
If you are upgrading from WOVN.php 1.1.1 or older, please manually add the following lines to your `.htaccess` files:
```

README.en.md
```diff
@@ -494,3 +500,43 @@ as possible, we would need to know information like the followings.
...
+<IfModule mod_rewrite.c>
+  RewriteEngine On
+
+  # Intercept only static content: html and htm urls
+  # Warning: do not remove this line or other content could be loaded
+  RewriteCond %{REQUEST_URI} /$ [OR]
+  RewriteCond %{REQUEST_URI} \.(html|htm|shtml|php|php3|phtml)
+  RewriteCond %{QUERY_STRING} !^off_wovn_php=1$
+  # Use the wovn_index.php to handle static pages
+  RewriteRule .? wovn_index.php [L]
```
a-bougouin, line 536: This section of the `.htaccess` is meant only for static websites. I guess you are adding it for `RewriteCond %{QUERY_STRING} !^off_wovn_php=1$` but it will have unwanted impact on PHP sites.

_Some history: Because it is difficult to fully support any website configuration with `.htaccess`, we deprecated wovnphp and created WOVN.php with two modes: a PHP mode where our script must be required from the PHP code that limits bugs related to content discovery (because we rely on devs rather than our `.htaccess`), and a static mode for non-PHP websites that relies on `.htaccess`_.

bstewart00 (approved):

Soichiro725 (approved):

src/wovnio/wovnphp/Utils.php
```diff
@@ -153,6 +153,20 @@ private static function checkIgnoreRegex($uri, $store = null)
         return false;
     }
 
+    public static function wovnDiagnosticsEnabled($store, $header)
+    {
+        if (strpos($header->url, 'wovn_diagnostics=1') == false) {
```
a-bougouin, line 158: Nitpick: if the diagnostic is enabled by query, then could we use `$header->query` instead of `$header->url`?

src/wovnio/utils/hash_gen.php
```diff
@@ -0,0 +1,7 @@
+<?php
```
a-bougouin, line 1: Was it a mistake to upload this file? It seems ununsed.

zeyuwu: Nope. This is a command line utility to generate the hash needed to login. If the customers need to login themselves they need this tool as well. More details on how usage can be found [here](https://wovnio.atlassian.net/browse/ST-277?focusedCommentId=17815).

src/wovnio/wovnphp/Utils.php
zeyuwu: Good point.

src/wovnio/utils/hash_gen.php
a-bougouin: I understand now, thanks. Is it gonna be documented somewhere else?

zeyuwu: Because this is considered internal operation procedures, I'm thinking adding a section on the Git Wiki in one of our private repositories, or a new confluence page.

src/wovnio/wovnphp/Utils.php
zeyuwu: Did some investigation, the `$header->query` is not exactly a representation of the real query (we processed it right after header creation, and only keep what we specify in the wovn.ini settings). So I guess `$header->url` is the best way to do it.

a-bougouin: I see. I guess you are talking about this: https://github.com/WOVNio/WOVN.php/blob/e122cf0b90247e3de7bff50947bb2d47e9f8ed9d/src/wovnio/wovnphp/Headers.php#L79.
Then I agree tat using `$header->url`.

a-bougouin (approved):

PR #121, opened 2020-08-27
Fixed incorrect handling of custom default lang alias.

zeyuwu:
This PR fixed the incorrect handling of custom default lang alias, resulting in malformed href links being generated when the `path` or `subdomian` pattern is used.

### Comments
In addition to bug fixes, this PR also:
1. added more testing related to custom default lang alas
2. refactored code into smaller, more streamlined modules for easier maintenance in the future.

Fixed the incorrect handling of custom default lang alias


src/wovnio/wovnphp/Url.php
```diff
@@ -49,89 +45,92 @@ public static function addLangCode($uri, $store, $lang, $headers)
...
-                        }
-                        break;
-                    case 'query':
-                        $new_uri = self::addQueryLangCode($no_lang_uri, $lang_code, $lang_param_name);
-                        break;
-                    default:
-                        //path
+            switch ($pattern) {
+                case 'subdomain':
+                    $lang_url = $headers->protocol . '://' . strtolower($lang_code) . '.' . $headers->host;
```
bstewart00, line 84: What is this code trying to do? I have no idea why subdomain pattern is dealing with paths at all.

zeyuwu: Didn't write this block of code, I only refactored them so they are less spaghetti-y. I think this block is trying to reconstruct a relative path back to a fully qualified path (protocol+host+path), then do the subdomain pattern thing on top.

src/wovnio/wovnphp/Url.php
```diff
@@ -150,21 +149,64 @@ private static function addQueryLangCode($uri, $lang_code, $lang_param_name)
...
+     * @param $no_lang_uri String URI without language code
+     * @return string|string[]|null
      */
-    public static function removeLangCode($uri, $lang_code, $settings)
+    private static function addSubdomainLangCode($parsed_url, $lang_code, $no_lang_uri)
     {
-        if (!$lang_code || strlen($lang_code) == 0) {
-            return $uri;
+        // check if subdomain already exists
+        if (preg_match('/.*\..*\..*/', $parsed_url['host'])) {
```
bstewart00, line 183: Not too sure about this one. Is this making assumptions about how many subdomains there are? `a.b.c.d.com` is valid, for example. Maybe look at how we do it in html-swapper and other libraries.

src/wovnio/wovnphp/Headers.php
```diff
@@ -63,15 +63,20 @@ public function __construct(&$env, &$store)
         $this->unmaskedUrl = $this->protocol . '://' . $this->unmaskedHost . $this->unmaskedPathname;
         $this->host = $this->unmaskedHost;
         if ($store->settings['url_pattern_name'] === 'subdomain') {
-            $this->host = $this->removeLang($this->host, $this->lang());
+            $intermediateHost = explode('//', $this->removeLang($this->protocol . '://' . $this->host, $this->lang()));
```
bstewart00, line 66: What was the motivation behind adding protocol to it? So `removeLang` can always assume the parameter is a full URI? I notice below this we call removeLang with just the query, though.

src/wovnio/wovnphp/Url.php
zeyuwu: That's a good point, totally also didn't consider the case where people having `company.co.uk` kind of domains. Will fix.

src/wovnio/wovnphp/Headers.php
zeyuwu: Yep, this is to fix a bug I discovered related to super **malformed href links when using subdomain pattern**. The remove_lang logic can only correctly handle subdomain patterns when the input is in the `http://english.example.com` form. I tried to find a root cause for this behavior but couldn't. Using this as a temporary fix I guess.

src/wovnio/wovnphp/Url.php
bstewart00: If we can prove this code doesn't actually do anything useful, it would be great to remove it. There really should not be much involved with add_lang_code for subdomain - add the lang code immediately after the protocol or before the first domain segment.

zeyuwu: My understanding is that `addLangCode` is trying to also account for cases when the input URI doesn't have its lang code removed. Maybe this is a behavior from older version where we don't explicitly "clean" the URIs. I will investigate this and if this behavior is no longer required, I'll remove the redundant code.

bstewart00: For a future refactor though I agree. Just shocked at seeing some of the weird stuff the code is doing 😒

zeyuwu: Yep, we definitely can straighten things up more. Will resolve this convo for now.

src/wovnio/wovnphp/Headers.php
```diff
@@ -63,15 +63,20 @@ public function __construct(&$env, &$store)
...
         }
         if ($store->settings['use_proxy'] && isset($env['HTTP_X_FORWARDED_REQUEST_URI'])) {
             $clientRequestUri = $env['HTTP_X_FORWARDED_REQUEST_URI'];
         } else {
             $clientRequestUri = $env['REQUEST_URI'];
         }
         $exploded = explode('?', $clientRequestUri);
-        $this->pathname = $this->removeLang($exploded[0], $this->lang());
+        if ($store->settings['url_pattern_name'] === 'subdomain') {
+            $this->pathname = $exploded[0];
```
bstewart00, line 76: Can't say I know the old code was doing here either. Is `$clientRequestUri` the path or full URL? Because I would expect `explode[0]` to include the host as well.

zeyuwu: It is always the path, like `/nice.html` or `/books/contemporary/fiction/`. The reason subdomain pattern is handled here differently is that we know for sure the path doesn't contain the lang when it's using subdomain pattern. The original logic screws up the path because the removeLang() can't handle a subdomain pattern without a full `scheme+host+path` combo.

bstewart00: Wish the code made this more obvious 😅  When I see URI i'm going to assume it is a URI

test/unit/html/HtmlConverterTest.php
```diff
@@ -108,6 +108,280 @@ public function testinsertSnippetAndHreflangTags()
...
+        $settings = array(
+            'supported_langs' => array('ja'),
+            'default_lang' => 'en',
+            'lang_param_name' => 'wovn',
+            'url_pattern_name' => 'path'
+        );
+        list($store, $headers) = StoreAndHeadersFactory::fromFixture('default', $settings);
+        $converter = new HtmlConverter($html, 'UTF-8', $store->settings['project_token'], $store, $headers);
+        $expected_href = 'http://my-site.com/ja/';
+        $generated_href = $this->invokeMethod($converter, 'buildHrefLang', array('ja'));
```
bstewart00, line 159: I don't really agree with testing private methods. Is this really the easiest way to test it right now?

zeyuwu: I'm doing this mostly to bring some level of confidence into the implementation. The chain of private method calls are too long before arriving at the public method making me feel it is not quite enough to just test the public ones. Also the private methods often have multiple behaviors, without testing them directly it's hard to figure out which behavior went wrong when the code breaks.

bstewart00: That's more of a problem with the class itself being too big, but tests are better than no tests, certainly. 

For this case I would just be calling the convert method and checking that the resulting HTML contains the expected hreflangs, for example.

zeyuwu: Fair point. I'll move the individual tests on private methods to `insertSnippetAndHreflangTags`.

bstewart00 (approved):

src/wovnio/wovnphp/Store.php
```diff
@@ -260,4 +260,15 @@ public function defaultLang()
...
         return $this->settings['default_lang'];
     }
+
+    /**
+     * Tests if a lang code is a valid lang code or lang alias.
+     *
+     * @param $code String the lang code to be tested
+     * @return bool if code is a valid Lang or Lang alias
+     */
+    public function isLangOrLangAlias($code)
```
Soichiro725, line 270: I could't find out the place where this method is used.

test/unit/html/HtmlConverterTest.php
```diff
@@ -108,6 +108,280 @@ public function testinsertSnippetAndHreflangTags()
         $this->assertEquals($expected_html, $translated_html);
     }
 
+    public function testinsertSnippetAndHreflangTagsWithCustomAlias()
```
Soichiro725, line 111: It should be capital I.

test/unit/html/HtmlConverterTest.php
```diff
@@ -108,6 +108,280 @@ public function testinsertSnippetAndHreflangTags()
...
+    /**
+     * Call protected/private method of a class.
+     *
+     * @param object &$object    Instantiated object that we will run method on.
+     * @param string $methodName Method name to call
+     * @param array  $parameters Array of parameters to pass into method.
+     *
+     * @return mixed Method return.
+     */
+    public function invokeMethod(&$object, $methodName, array $parameters = array())
```
Soichiro725, line 138: It is easy to use this method, if this function is in the test helper.

zeyuwu: Thank you. I'll replace them.

src/wovnio/wovnphp/Store.php
zeyuwu: Sorry totally missed this one. I added this one thinking it might be useful during the refactoring process, but didn't end up using it. I'll remove this one.

Soichiro725 (approved):

PR #122, opened 2020-09-03
Refactoring/the big change

zeyuwu:
This PR added the frame of a new core library for developing WOVN.io libraries in PHP. The current capability of the core library includes:
- loading and parsing of user configuration
- detecting url language

### Comments
Started a new master branch for future cumulative updates for the core library.

Both Wovn PHP Core and it's integration into WOVN.php will be parked in this new branch at the moment.

Added the frame of a new core library for developing WOVN.io libraries in PHP.


zeyuwu:
I'm a little bummed out by the warnings. Sigh.

src/wovnio/wovnphp/core/WovnLang.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * WovnLang constructor.
+     * @param $code string The language code, IETF language tag format.
+     * @param $name string The name of the language, in its local language.
+     * @param $englishName string The name of the language, in English.
+     * @param mixed $alias The alias of this language, defaults to null.
+     * @param bool $active If the language is a valid choice in the current project.
+     */
+    public function __construct($code, $name, $englishName, $alias = null, $active = false)
```
bstewart00, line 22: I would say this class is doing too much personally. Don't really like the "active" status, and I would make language aliases be managed separately from this class myself.

Methods like "setAlias", "enable" are kind of off. These things should only be getting set once on initialization, so don't really deserve methods for.

src/wovnio/wovnphp/core/WovnLang.php
```diff
@@ -0,0 +1,114 @@
...
+     */
+    public function setAlias($alias)
+    {
+        $this->alias = $alias;
+    }
+
+    /**
+     * Marks the language as in-use in the project.
+     */
+    public function enable()
```
bstewart00, line 100: make this `setActive` if you're going to have it. Currently you need to read the comment to know what it does

src/wovnio/wovnphp/core/url_handlers/WovnPathURLHandler.php
```diff
@@ -0,0 +1,54 @@
...
+            if (!$lang->isValidLang()) {
+                $lang = $this->langDirectory->defaultLang();
+            }
+        } catch (WovnLangException $e) {
+            $lang = $this->langDirectory->defaultLang();
+        }
+        return $lang;
+    }
+
+    protected function removeLang()
```
bstewart00, line 40: This is OK to leave as stubs for now, but note that the `addLang/removeLang` approach is kind of  out of date after you take in account newer features like default language aliases, custom domain. 
Sometimes you need to "add" a code to the URL to arrive at the default language. 

We would do well to move towards `convertToDefaultLang` and `convertToTargetLang` which are the more higher level operations we actually mean.

src/wovnio/wovnphp/core/tests/unit/WovnURLTest.php
```diff
@@ -0,0 +1,262 @@
...
+require_once 'src/wovnio/wovnphp/core/url_handlers/WovnQueryURLHandler.php';
+require_once 'src/wovnio/wovnphp/core/url_handlers/WovnPathURLHandler.php';
+require_once 'src/wovnio/wovnphp/core/url_handlers/WovnSubdomainURLHandler.php';
+require_once 'src/wovnio/wovnphp/core/WovnURL.php';
+
+use Wovnio\Wovnphp\Core\WovnLangDirectory;
+use Wovnio\Wovnphp\Core\WovnOption;
+use Wovnio\Wovnphp\Core\WovnURL;
+
+class WovnURLTest extends \PHPUnit_Framework_TestCase
```
bstewart00, line 21: I don't see a purpose for WovnURL at the moment, versus just having WovnUrlHandler with potentially an abstract base class for common logic. At the moment it isn't actually doing anything except delegate to WovnURL.

Ideally, as far as tests go we should be testing the url handlers individually. Having all patterns in a single file makes it so messy.

src/wovnio/wovnphp/core/tests/unit/WovnURLTest.php
```diff
@@ -0,0 +1,262 @@
...
+    public function testConstructorWithPathPatternNonDefaultLangWithPathPrefixVariation2()
+    {
+        $directory = $this->getLangDirectory();
+        $options = $this->getOptions(array(WovnOption::OPT_URL_PATTERN_NAME => 'path', WovnOption::OPT_SITE_PREFIX_PATH => 'news/blog'));
+        $originalUrl = 'https://www.example.com/news/blog/ja/text/next/page.html';
+        $url = new WovnURL($originalUrl, $directory, $options);
+        self::assertEquals('ja', $url->lang()->code());
+    }
+
+    public function testConstructorWithPathPatternNonDefaultLangWithPathPrefixVariation3()
```
bstewart00, line 245: For these test names, i'm seeing `Constructor` a lot but what you are actually testing is the `lang` function.

src/wovnio/wovnphp/core/WovnLang.php
zeyuwu: Yep, I agree the setAlias() and enable() should not be there. These were hacked on when I was working on LangDirectory.

But I still believe it's necessary to keep information such as if the language is in use in the project or the language's alias inside the Lang class. My rationale is as the following:

1. Logically the alias of a language and its validity are information that describes a language
2. It is easier to do `$lang->alias()` than `LangAlias->getAlias($lang)`, like the focus is not broken.

bstewart00: It is definitely more convenient, but I think it comes at the expense of code "purity" 😛 

Another operation we need to do is reverse-lookup a language from an alias, and I think there is symmetry with having both of these operations being close together in code.

bstewart00:
We don't need the Wovn prefix on everything IMO. We also use namespaces already.

bstewart00:
This is obviously a prototype so we don't need to get the design perfect straight away. I just want to be cautious that we don't repeat mistakes from wovn.php.

I see the base branch is `refactoring_master`, so I am OK with committing this as a starting point if you want.

src/wovnio/wovnphp/core/WovnLang.php
zeyuwu: The reverse lookup is actually handled by "getting" the lang. The design is that whenever you want to get a Lang object, you call `WovnLangDirectory->getLang()`, it doesn't matter if you pass in a proper lang code or an alias. And you also get the same Lang object back for the same language.

bstewart00 (approved):

src/wovnio/wovnphp/core/WovnLang.php
bstewart00: Yeah, I get that WovnLangDirectory is kind of serving both purposes.

src/wovnio/wovnphp/core/WovnRequest.php
```diff
@@ -0,0 +1,92 @@
...
+            $this->options->get(WovnOption::OPT_CUSTOM_LANG_ALIASES)
+        );
+        $this->wovnUrl = new WovnURL($this->url, $this->langDirectory, $this->options);
+        $this->lang = $this->wovnUrl->lang();
+    }
+
+    /**
+     * Swaps the HTML either locally or via API, depending on the situation
+     */
+    public function swap()
```
bstewart00, line 45: i'm not sure what functionality is planned to be in here, but it feels beyond the scope of this class. It would be better to keep this class just related to telling you about the state of the current request, what language has been calculated for it etc.

Taking the response body/headers and translating it should be done by something else.

zeyuwu: The WovnRequest class is supposed to be the "entry point" for integration, as this will allow the maximum amount of abstraction (you only need to provide a $_SERVER superglobal and the wovn configuration). And the swap() function is exposed to the developers as the only public function (right now lang() and wovnUrl() are exposed to help with partial integration). The idea is to just instantiate WovnRequest object and call the swap() function, then you get a translated or original HTML, with the server headers already modified for you, ready to be sent out.

bstewart00: OK, well I would tend more towards a name like `translate`.  "swapping" is just one part of what these libraries do

src/wovnio/wovnphp/core/WovnRequest.php
```diff
@@ -0,0 +1,92 @@
...
+    private $options;
+
+    public function __construct($serverSuperGlobal, $optionConfig, $useStrictMode = true)
+    {
+        $this->options = new WovnOption($optionConfig, $useStrictMode); // Do Not handle exceptions here, let them fail.
+        $this->scheme = $this->parseScheme($serverSuperGlobal);
+        $this->host = $serverSuperGlobal['HTTP_HOST'];
+        $this->path = $this->parseURI($serverSuperGlobal);
+        $this->query = isset($serverSuperGlobal['QUERY_STRING']) ? $serverSuperGlobal['QUERY_STRING'] : '';
+        $this->url = $this->scheme . '://' . $this->host . $this->path . $this->query;
```
Soichiro725, line 32: When I access with `http://testsite.com/?wovn=en`, $this->url will be `http://testsite.com/?wovn=enwovn=en`.

zeyuwu: Thank you for pointing this out. I didn't realize the REQUEST_URI includes the query string.

Soichiro725 (approved):

PR #126, opened 2020-09-15
[Proof of concept]Custom Domain Langs feature

Soichiro725:
This is the proof of concept about `custom domain langs` feature.

### DIFF
--- a/src/wovnio/wovnphp/Headers.php
+++ b/src/wovnio/wovnphp/Headers.php
@@ -191,23 +191,32 @@ class Headers
             } else {
                 $server_name = $this->env['SERVER_NAME'];
             }
-            // get the lang in the path
-            $rp = '/' . $this->store->settings['url_pattern_reg'] . '/';
+
             if ($this->store->settings['use_proxy'] && isset($this->env['HTTP_X_FORWARDED_REQUEST_URI'])) {
                 $request_uri = $this->env['HTTP_X_FORWARDED_REQUEST_URI'];
             } else {
                 $request_uri = $this->env['REQUEST_URI'];
             }
-            preg_match($rp, $server_name . $request_uri, $match);
-            if (isset($match['lang'])) {
-                $lang_code = Lang::formatLangCode($match['lang'], $this->store);
-                if (!is_null($lang_code)) {
-                    $this->pathLang = $lang_code;
+
+            $url = $server_name . $request_uri;
+            $lang_identifier = null;
+            // get the lang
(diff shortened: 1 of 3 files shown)


src/wovnio/wovnphp/Headers.php
```diff
@@ -191,23 +191,32 @@ public function computePathLang()
...
-                $lang_code = Lang::formatLangCode($match['lang'], $this->store);
-                if (!is_null($lang_code)) {
-                    $this->pathLang = $lang_code;
+
+            $url = $server_name . $request_uri;
+            $lang_identifier = null;
+            // get the lang
+            if ($this->store->settings['url_pattern_name'] == 'custom_domain') {
+                foreach ($this->store->settings['custom_domain_langs'] as $lang_url => $lang) {
+                    if (strpos($url, $lang_url) !== false) {
```
Soichiro725, line 206: When judge current lang from URL, we need more restrict check.

e.g.
```
{
  custom_domain_langs: { testsite.com: 'ja', testsite.com/en: 'en'},
  default_lang: 'ja' 
}
```
Case:
  - testsite.com/enxxx/index.html -> ja
  - testsite.com/en.index.html -> ja
  - testsite.com/en/index.html -> en

Tasks:
- Order custom_domain_langs rules from longer one, before use it
- Add `/`, if path doesn't have it

src/wovnio/wovnphp/Headers.php
```diff
@@ -392,6 +401,17 @@ public function removeLang($uri, $lang = null)
...
         $lang_code = $this->store->convertToCustomLangCode($lang);
         $default_lang = $this->store->settings['default_lang'];
+        $url_pattern_name = $this->store->settings['url_pattern_name'];
+        $custom_domain_langs = $this->store->settings['custom_domain_langs'];
+        if ($url_pattern_name == 'custom_domain' && isset($custom_domain_langs)) {
+            $current_lang_domain = array_search($lang_code, $custom_domain_langs);
+            $default_lang_domain = array_search($default_lang, $custom_domain_langs);
+            $current_lang_path = parse_url('http://' . $current_lang_domain . '/')['path'];
+            $default_lang_path = parse_url('http://' . $default_lang_domain . '/')['path'];
+            return str_replace($current_lang_path, $default_lang_path, $uri);
```
Soichiro725, line 411: To generate original HTML path, path needs to be generated from custom_domain_langs.
But custom_domain_langs's url includes host.
So, host part should be removed first.

Soichiro725:
The `custom domain langs` feature has already been implemented.
I will erase this PR because this is no longer needed.

PR #127, opened 2020-09-25
[Feature]Custom domain langs

Soichiro725:
https://wovnio.atlassian.net/browse/ST-486

Add `custom domain langs` feature.


src/wovnio/wovnphp/Url.php
```diff
@@ -238,6 +251,28 @@ public static function shouldIgnoreBySitePrefixPath($uri, $settings)
         return false;
     }
 
+    public static function parseUrlStr($url)
```
bstewart00, line 254: Why aren't you using `parse_url` here?

Soichiro725: I want to use parse_url, but parse_url is not enough.
When parse `testsite.com` with parse_url, path will be `testsite.com`.
So, I needed to wrap parse_url.

src/wovnio/wovnphp/Headers.php
```diff
@@ -392,6 +402,15 @@ public function removeLang($uri, $lang = null)
...
         $lang_code = $this->store->convertToCustomLangCode($lang);
         $default_lang = $this->store->settings['default_lang'];
+        $url_pattern_name = $this->store->settings['url_pattern_name'];
+        $custom_domain_langs = $this->store->settings['custom_domain_langs'];
+        if ($url_pattern_name == 'custom_domain' && isset($custom_domain_langs)) {
+            $current_lang_domain = array_search($lang_code, $custom_domain_langs);
+            $default_lang_domain = array_search($default_lang, $custom_domain_langs);
+            $current_lang_path = URL::parseUrlStr($current_lang_domain)['path'];
+            $default_lang_path = URL::parseUrlStr($default_lang_domain)['path'];
+            return str_replace($current_lang_path, $default_lang_path, $uri);
```
bstewart00, line 412: Look at https://github.com/WOVNio/Wovn.cs/blob/master/wovn.cs/Helpers/URL.cs#L109

This should be using the same function as addLangCode. This looks like it is duplicated. 

Do not be afraid to create individual functions for custom domain. I would like to see a similar `ChangeToNewCustomDomainLanguage` function.

src/wovnio/wovnphp/Url.php
bstewart00: I'm not sure what you mean.
```
print_r(parse_url("http://test.com/dir/page.html"));

Array
(
    [scheme] => http
    [host] => test.com
    [path] => /dir/page.html
)
```

Soichiro725: I mean that `json_encode(parse_url('testsite.com')); ` will be `{"path":"testsite.com"}`.

src/wovnio/wovnphp/Headers.php
```diff
@@ -198,16 +198,26 @@ public function computePathLang()
...
             }
-            preg_match($rp, $server_name . $request_uri, $match);
-            if (isset($match['lang'])) {
-                $lang_code = Lang::formatLangCode($match['lang'], $this->store);
-                if (!is_null($lang_code)) {
-                    $this->pathLang = $lang_code;
+
+            $full_url = $server_name . $request_uri;
+            $lang_identifier = null;
+            if ($this->store->settings['url_pattern_name'] == 'custom_domain') {
```
bstewart00, line 204: Do not put all of this code in Headers. Move it to a custom_domain.php file at least. Please keep in mind how it is done in Wovn.cs. I am not sure if this is correct. https://github.com/WOVNio/Wovn.cs/blob/master/wovn.cs/CustomDomainLanguages/CustomDomainLanguagesModel.cs#L22

src/wovnio/wovnphp/Url.php
bstewart00: When are you encountering URLs that do not have the protocol? We should be doing additional work in wovn.php to ensure the URL is in a consistent format.

src/wovnio/wovnphp/Headers.php
```diff
@@ -198,16 +198,26 @@ public function computePathLang()
...
+                foreach ($this->store->settings['custom_domain_langs'] as $lang_url => $lang) {
+                    if (strpos($full_url, $lang_url) !== false) {
```
zeyuwu, line 206: Is this safe? I know this is an unlikely scenario but what if we have setting `['custom_domain_langs'] = {"test.com"}`, but the url is `https://test.com.something.else.com/index.php`?

src/wovnio/wovnphp/Headers.php
Soichiro725: I'm trying to use removeLangCode in Url.php from Headers.php.
And I will replace url host and path separately.

src/wovnio/wovnphp/Url.php
bstewart00: If we do not have access to the full URL, how do we expect to implement custom domain properly? We need to know if the incoming request is coming from "ja.site.com/foo" vs "en.site.com"

bstewart00: This should be similar to the subdomain pattern where it uses HTTP_HOST. I understand HTTP_HOST may not always be available depending on configuration, in that case we should fallback to the default custom domain host.

src/wovnio/wovnphp/Headers.php
bstewart00: I agree. Please try to stay as close as you can to the wovn.cs implementation. We should be checking the host component against all of the custom domain languages, followed by the path.

Soichiro725: I will review wovn.cs and copy as much as possible.

bstewart00:
merge with master please, I think you are using some code that has been deleted.

src/wovnio/wovnphp/Url.php
```diff
@@ -99,6 +104,14 @@ public static function addLangCode($uri, $store, $lang, $headers)
                 case 'query':
                     $new_uri = self::addQueryLangCode($no_lang_uri, $lang_code, $lang_param_name);
                     break;
+                case 'custom_domain':
+                    if (self::isAbsolutePath($no_lang_uri)) { // absolute path
```
bstewart00, line 108: WOVN.php relative path handling is pretty broken so this is OK for now. The correct way would be converting relative paths to absolute, but this is likely not easy in PHP.

src/wovnio/wovnphp/custom_domain/CustomDomainLang.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $host = $parsedUrl['host'];
+        $path = array_key_exists('path', $parsedUrl) ? $parsedUrl['path'] : '/';
+        return strtolower($host) === strtolower($this->host) && $this->pathIsEqualOrSubsetOf($this->path, $path);
+    }
+
+    private function pathIsEqualOrSubsetOf($path1, $path2)
+    {
+        // split by delimiter and remove spaces and empty strings
+        $path1Segments = array_filter(array_map('trim', explode($this->delimiter, $path1)), 'strlen');
```
bstewart00, line 43: just remove the delimiter variable and use '/' directly

src/wovnio/wovnphp/Headers.php
```diff
@@ -320,6 +327,23 @@ public function requestOut()
                 $this->env['SERVER_NAME'] = $this->removeLang($this->env['SERVER_NAME']);
                 break;
 
+            case 'custom_domain':
```
zeyuwu, line 330: Just something nice to have: in requestOut() there are a lot of duplicate code fragments, maybe we can extract them out like so:

    private function uriRemoveLang()
    {
        if (isset($this->env['REQUEST_URI'])) {
            $this->env['REQUEST_URI'] = $this->removeLang($this->env['REQUEST_URI']);
        }
        if (isset($this->env['REDIRECT_URL'])) {
            $this->env['REDIRECT_URL'] = $this->removeLang($this->env['REDIRECT_URL']);
        }
        if (isset($this->env['HTTP_X_FORWARDED_REQUEST_URI'])) {
            $this->env['HTTP_X_FORWARDED_REQUEST_URI'] = $this->removeLang($this->env['HTTP_X_FORWARDED_REQUEST_URI']);
        }
    }

    private function hostRemoveLang()
    {
        if ($this->store->settings['use_proxy'] && isset($this->env['HTTP_X_FORWARDED_HOST'])) {
            $this->env['HTTP_X_FORWARDED_HOST'] = $this->removeLang($this->env['HTTP_X_FORWARDED_HOST']);
        }
        $this->env['HTTP_HOST'] = $this->removeLang($this->env['HTTP_HOST']);
        $this->env['SERVER_NAME'] = $this->removeLang($this->env['SERVER_NAME']);
    }

src/wovnio/wovnphp/Url.php
```diff
@@ -224,6 +247,26 @@ public static function removeLangCode($uri, $lang_code, $settings)
...
+                $newUri = $uri;
+                if (self::isAbsoluteUri($uri)) {
+                    $newUri = CustomDomainLangUrlHandler::changeToNewCustomDomainLang($uri, $customDomainLangToRemove, $defaultCustomDomainLang);
+                } elseif (self::isAbsolutePath($uri)) {
+                    $absoluteUrl = $headers->protocol . '://' . $headers->unmaskedHost . $uri;
+                    $absoluteUrlWithLang = CustomDomainLangUrlHandler::changeToNewCustomDomainLang($absoluteUrl, $customDomainLangToRemove, $defaultCustomDomainLang);
+                    $segments = self::makeSegmentsFromAbsoluteUrl($absoluteUrlWithLang);
+                    $newUri = $segments['others'];
+                } elseif ($uri === $customDomainLangToRemove->getHost()) {
+                    $absoluteUrl = $headers->protocol . '://' . $uri . $headers->unmaskedPathname;
```
bstewart00, line 264: We need `originalHost` and `originalPath` for these cases. The $pathname, $host variables in Headers will have the language removed already, but to remove the language from `french.testsite.com/fr/`, we need to combine `french.testsite.com` with `/fr/`, not `/`

src/wovnio/wovnphp/Headers.php
bstewart00: agree, I was kind of expecting we would have `removeLangFromPath`, `removeLangFromHost` functions

bstewart00:
At the moment it doesn't work for French with 
```
url_pattern_name = custom_domain
default_lang = en
supported_langs[] = ja
supported_langs[] = fr
encoding = UTF-8
custom_domain_langs[localphp.com] = en
custom_domain_langs[localphp.com/ja] = ja
custom_domain_langs[french.localphp.com/fr] = fr

```
but still trying to find out why.

src/wovnio/wovnphp/Url.php
bstewart00: actually, looks like $unmaskedPath and $unmaskedHost do serve this purpos. Once you merge with master you may need to re-add them. Rename them to `originalHost` etc though.

src/wovnio/wovnphp/Headers.php
```diff
@@ -62,7 +62,7 @@ public function __construct(&$env, &$store)
         }
         $this->unmaskedUrl = $this->protocol . '://' . $this->unmaskedHost . $this->unmaskedPathname;
         $this->host = $this->unmaskedHost;
-        if ($store->settings['url_pattern_name'] === 'subdomain') {
+        if (in_array($store->settings['url_pattern_name'], array('subdomain', 'custom_domain'))) {
             $intermediateHost = explode('//', $this->removeLang($this->protocol . '://' . $this->host, $this->lang()));
```
bstewart00, line 66: This is not quite good enough. Since this URL will appear as an absolute URL with no path `http://french.testsite.com`, removeLang will not do anything. For custom domain we also need to append the request path in order to remove lang from the host.

bstewart00: Alternatively, wouldn't changing this to `$this->removeLang($this->host, $this->lang()));` be OK? subdomain pattern does not care about the protocol anyway.

zeyuwu:
Did some tests locally I think it works, well enough?

Soichiro725:
> At the moment it doesn't work for French with
> 
> ```
> url_pattern_name = custom_domain
> default_lang = en
> supported_langs[] = ja
> supported_langs[] = fr
> encoding = UTF-8
> custom_domain_langs[localphp.com] = en
> custom_domain_langs[localphp.com/ja] = ja
> custom_domain_langs[french.localphp.com/fr] = fr
> ```
> 
> but still trying to find out why.

With this config, WOVN.php works for my local environemnt.

zeyuwu:
Weird. I had almost the same config, and even before your update it works for me.
```

custom_domain_langs[testsite.com/ja] = "en"
custom_domain_langs[testsite.com/actual_ja] = "ja"
custom_domain_langs[french.testsite.com] = "fr"
```

bstewart00:
@zeyuwu the `/fr/` is important 😉

src/wovnio/wovnphp/Headers.php
zeyuwu: I don't remember all the details but I too discovered that subdomain patterns specifically requires this special handling, or it wouldn't work.

Soichiro725:
I added test/integration/UrlCustomDomainPatternTest.php for integration test.
And, I started to check `$_SERVER` variables, after lang removed.
By this tests, we can test removeLang works.

src/wovnio/wovnphp/Headers.php
Soichiro725: `addLangCode` method in Url.php will add protocol with subdomain pattern.
And after adding lang code, it doesn't remove protocol 😢

bstewart00:
My french config is working with the latest changes.

zeyuwu (approved):
Looks good.

src/wovnio/wovnphp/Url.php
```diff
@@ -243,7 +286,7 @@ private static function isAnchorLink($uri)
         return preg_match('/^(#.*)?$/', $uri);
     }
 
-    private static function generateUrlRegex($site_prefix_path = '')
+    public static function generateUrlRegex($site_prefix_path = '')
```
bstewart00, line 289: I don't think this should be public. Nothing else is using it anyway.

Soichiro725:
I added some documents to `HOWTO.md` .
You can run unit tests and integration tests with PHP version which you like.
This way is easy and fast to run.

bstewart00 (approved):

zeyuwu:
> You can run unit tests and integration tests with PHP version which you like.

Should we stop using the old `./scripts/run_phpunit_tests.sh php:7.3-apache` ?

Soichiro725:
> > You can run unit tests and integration tests with PHP version which you like.
> 
> Should we stop using the old `./scripts/run_phpunit_tests.sh php:7.3-apache` ?

We can continue to use it.
But this way is slow to run tests, because it needs to preparing docker env.
We can prepare docker env only once and run tests directly by using test.yml.

PR #128, opened 2020-09-29
Changed to rely on content-type/mime-type to determine if the buffer …

zeyuwu:
…should be translated.

https://wovnio.atlassian.net/browse/ST-324

### Comments
The general idea is that we can use the mime-type of a file to determine if it should be translated (html etc) or not, instead of relying on looking at the request URI's file extension.

Removed regex based file extention matcher competely. Requires the client to have `ext-fileinfo` extension.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,8 @@
   "description": "PHP server library for WOVNio",
   "homepage": "https://wovn.io",
   "require": {
-    "php": ">=5.3"
+    "php": ">=5.3",
+    "ext-fileinfo": "*"
   },
   "require-dev": {
     "phpunit/phpunit": "4.3.*",
--- a/src/wovn_interceptor.php
+++ b/src/wovn_interceptor.php
@@ -34,12 +34,6 @@ $headers->requestOut();
 
 $uri = $headers->getDocumentURI();
 if (!Utils::isIgnoredPath($uri, $store)) {
-    $filePath = wovn_helper_detect_paths(__DIR__ . '/../..', $uri);
-    $mimeType = false;
-    if (!empty($filePath)) {
-        $mimeType = mime_content_type($filePath[0]);
-    }
-
     $diagnostics = null;
     $benchmarkStart = 0;
     if (Utils::wovnDiagnosticsEnabled($store, $headers)) {
@@ -48,11 +42,11 @@ if (!Utils::isIgnoredPath($uri, $store)) {
         $diagnostics = new Diagnostics($store);
     }
     // use the callback of ob_start to modify the content and return
-    ob_start(function ($buffer) use ($headers, $store, $diagnostics, $benchmarkStart, $mimeType) {
+    ob_start(function ($buffer) use ($headers, $store, $diagnostics, $benchmarkStart) {
 
         $headers->responseOut();
 
-        if (empty($buffer) || !Utils::isHtml($mimeType, $buffer)) {
+        if (empty($buffer) || !Utils::isHtml($buffer)) {
             return $buffer;
         }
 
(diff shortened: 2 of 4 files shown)


src/wovn_interceptor.php
```diff
@@ -31,9 +33,13 @@
 $headers->requestOut();
 
 $uri = $headers->getDocumentURI();
-if (!Utils::isFilePathURI($uri, $store) &&
-    !Utils::isIgnoredPath($uri, $store)
-    ) {
+if (!Utils::isIgnoredPath($uri, $store)) {
+    $filePath = wovn_helper_detect_paths(__DIR__ . '/../..', $uri);
```
bstewart00, line 37: is this trying to find the file on disk that is coming from the URL, so we can detect the mime type?

zeyuwu: Yep. I reused the same logic from wovn helper for that.

Soichiro725, line 37: Does this code expect that WOVN.php is on the root dir?
I saw some clients are using WOVN.php with not root dir :cry:

zeyuwu: Yeah it does assume WOVN.php folder is at document root. Let me change it to make use of _SERVER['DOCUMENT_ROOT']

bstewart00:
I'm wondering if there is still value in keeping a response header check for Content-Type. Do we need to care about URLs like `/foo` don't map to a file but return a response with `application/pdf` etc? The buffer check will handle that I suppose?

zeyuwu:
Actually I think I can imporve this check to have it directly check the response buffer. That'll handle all cases.

src/wovn_interceptor.php
zeyuwu: Since we no longer check the request file, but only the response buffer, this is no longer a problem.

zeyuwu: Since we no longer check the request file, but only the response buffer, this is no longer a problem.

composer.json
```diff
@@ -4,7 +4,8 @@
   "description": "PHP server library for WOVNio",
   "homepage": "https://wovn.io",
   "require": {
-    "php": ">=5.3"
+    "php": ">=5.3",
+    "ext-fileinfo": "*"
```
a-bougouin, line 8: This is the first time we require a dependency for WOVN.php. I don't think any of our customer use composer for WOVN.php installation, so we need to make it clear. Please talk with Swappers and OnSite to decide the best strategy to adopt. In the long run, publishing WOVN.php so it can be installed from composer would be the best.

zeyuwu: This "dependency" is kind of special is that it is a PHP extension module that is shippped by default for PHP 5.3+, which also coincides with the PHP versions WOVN.php targets. So, strictly speaking no installation is required and the user needs to at most enable this feature. I think this can be done by adding a line in the installation instruction.

a-bougouin: I see, looks good, thanks.

bstewart00 (approved):

Soichiro725:
XML file like the followings cause parse issue at `wovn_helper_include_by_paths` method.

```
<?xml version="1.0" encoding="UTF-8" ?>
<foods>
  <food>
    <name>バナナ</name>
    <color>黄色</color>
  </food>
</foods>
```

XML file is judged as html by `isHtml` :cry:

Soichiro725:
But I guess we support xhtml.

zeyuwu:
> XML file like the following cause parse issue at wovn_helper_include_by_paths method.

For this I don't think it's a problem. The wovn_helper_include_by_paths is used by static sites, which would require the specific .htaccess rules that bypass anything that's not HTML or PHP.

zeyuwu:
I just tested a bit more, if the XML is returned as the result of PHP execution, then it is not recognized correctly. I'd get `text/xml; charset=utf-8`.

Soichiro725:
xhtml file in static site will cause issue.
But I guess this is minor problem. (I hope.)

Soichiro725 (approved):

a-bougouin (approved):

PR #130, opened 2020-10-01
Remove query feature

bstewart00:
### Comments
This refers to the `query[]=foo` setting that was apparently a thing. It was added on the very first commit of WOVN.php, but I am not convinced it is actually used. I am assuming this was carried over from the old wovnphp that did not use html swapper.

Equalizer _does_ have a query parameter feature, where you can specify which query parameters should be considered a "unique" page in our system. However, this is no longer any concern of backends. This logic is done in the `js_data` request during auto page add logic (the fact that the widget does not have any code for this feature is also telling). 

WOVN.php already sends the full URL including query params to HTML swapper (regardless of this setting). Then HTML swapper forwards this to the js_data request. So I am not sure that this setting actually _does_ anything right now other than creating a bunch of variables.

So my opinion is we can delete it. Please offer justification if you think we still need this.


zeyuwu (approved):
The Japanese README now have new newlines?

Soichiro725 (approved):

bstewart00:
> The Japanese README now have new newlines?

Looks like just trailing spaces were removed. Should be fine,

PR #131, opened 2020-10-02
Feature/allow ignore regex with query

zeyuwu:
To provide a fix for https://wovnio.atlassian.net/browse/SUP-355

### Comments
This PR allows ignore_regex to work on the query part of URLs too.
Also changed recommended way of setting wovn.ini (with double quotes around values)

### DIFF
--- a/src/wovnio/wovnphp/Headers.php
+++ b/src/wovnio/wovnphp/Headers.php
@@ -314,11 +314,7 @@ class Headers
         $url = $this->env['REQUEST_URI'];
         $url_arr = parse_url($url);
 
-        if ($withQuery) {
-            return $url;
-        }
-
-        if ($url_arr && array_key_exists('query', $url_arr)) {
+        if (!$withQuery && $url_arr && array_key_exists('query', $url_arr)) {
             $query = $url_arr['query'];
             $uri = str_replace(array($query,'?'), '', $url);
         } else {
--- a/src/wovnio/wovnphp/Headers.php
+++ b/src/wovnio/wovnphp/Headers.php
@@ -309,12 +309,12 @@ class Headers
         }
     }
 
-    public function getDocumentURI($withQuery = false)
+    public function getDocumentURI()
     {
         $url = $this->env['REQUEST_URI'];
         $url_arr = parse_url($url);
 
-        if (!$withQuery && $url_arr && array_key_exists('query', $url_arr)) {
+        if ($url_arr && array_key_exists('query', $url_arr)) {
             $query = $url_arr['query'];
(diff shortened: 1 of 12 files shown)


src/wovnio/wovnphp/Headers.php
```diff
@@ -309,11 +309,15 @@ public function removeLang($uri, $lang = null)
...
     }
 
-    public function getDocumentURI()
+    public function getDocumentURI($withQuery = false)
     {
         $url = $this->env['REQUEST_URI'];
         $url_arr = parse_url($url);
 
+        if ($withQuery) {
+            return $url;
```
Soichiro725, line 318: The following code is skipped.
```
return $this->removeLang($uri, $this->lang());
```

src/wovnio/wovnphp/Utils.php
```diff
@@ -97,11 +97,11 @@ public static function isFilePathURI($uri, $store)
      * Return true if $uri should be ignored according to `ignore_paths` or `ignore_regex`.
      * Return false otherwise.
      */
-    public static function isIgnoredPath($uri, $store)
```
Soichiro725, line 100: How about always use URL with query.

zeyuwu: Hmm I guess that's OK. Initially I didn't want to do it because I worry that ignore_paths might not behave correctly. But after looking at the code i think it's ok.

src/wovnio/wovnphp/Headers.php
bstewart00: Say no to boolean function parameters. I especially don't agree with using it just as a way to skip a good chunk of the method (that url parsing becomes useless).

I've been thinking of removing the `getDocumentURI` function entirely since it doesn't seem to be doing much. So I wish we weren't adding to it.

bstewart00: By the way, since I am removing the query feature (which seems to be what is blocking you here). There might not even be a need for this change?

zeyuwu: I don't think this is related to the query feature you are removing tho. It's just that in the past for some reason when pulling the URI to determine if it should be ignored, we chose to get rid of the queries in it.

bstewart00: Yes, in my PR I also removed this code because I deemed it useless 😛 Try merging with master and see if you actually need to make any changes.

zeyuwu: Ahh I see. Yep should work fine.

Soichiro725 (approved):

bstewart00 (approved):

PR #133, opened 2020-10-07
Feature/custom domain langs

zeyuwu:
https://wovnio.atlassian.net/browse/ST-502

Accompanying documentation is here: https://github.com/WOVNio/swappers/pull/58/files?short_path=85a444c#diff-85a444ce878a6ab77c13bf69733bff7641e5883bf31a6912d19ff0af9bae566b

### Comments
This PR brought the following changes to  WOVN.php:

1. WOVN.php now supports JSON format configurations files (default name: wovn.ini, default path: at root of document root, 2 levels up from inside wovnphp folder).
2. This JSON format is required for the new Custom Domain Langs to work.
3. Added a new feature to Custom Domain Langs. You can now specify a "source" attribute for custom domain lang settings. This source refers to the actual location of your document for that specific language.

Added new `source` attribute to custom domain settings.


src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
```diff
@@ -1,22 +1,39 @@
...
             $parsedUrl = parse_url($this->addProtocolIfNeeded($langUrl));
-
+            $source = array_key_exists('source', $config) ? $config['source'] : null;
             // Disable notice error by adding @, when path is not defined
-            array_push($this->customDomainLangs, new CustomDomainLang($parsedUrl['host'], @$parsedUrl['path'], $lang));
+            array_push($this->customDomainLangs, new CustomDomainLang($parsedUrl['host'], @$parsedUrl['path'], $config['lang'], $source));
         }
     }
 
+    public function getSourceCustomDomainByLang($langCode)
```
bstewart00, line 24: How i'm expecting this to work is:
 -Every language can have a source associated with it meaning where the raw HTML comes from
- if a given language does not have a source specified for it, default to the default language's source.

Would like to avoid "bypass" in our naming if possible. Physical URL would be preferred.

Also avoid using this->settings['custom_domain_langs']. Everything should be getting delegated through the CustomDomainLangs class.

src/wovnio/wovnphp/Utils.php
```diff
@@ -8,7 +8,10 @@ class Utils
     // will return the store and headers objects
     public static function getStoreAndHeaders(&$env)
     {
-        $file = isset($env['WOVN_CONFIG']) ? $env['WOVN_CONFIG'] : DIRNAME(__FILE__) . '/../../../../wovn.ini';
+        $file = isset($env['WOVN_CONFIG']) ? $env['WOVN_CONFIG'] : DIRNAME(__FILE__) . '/../../../../wovn.json';
+        if (!file_exists($file)) {
```
Soichiro725, line 12: How about using `WOVN_CONFIG`, when user needs to use json format.
Then, WOVN.php doesn't need to check wovn.json exist or not.

src/wovnio/wovnphp/Store.php
```diff
@@ -22,6 +22,12 @@ class Store
      */
     public static function createFromFile($settingFileName)
     {
+        if (strpos($settingFileName, "json")) {
```
Soichiro725, line 25: `substr('wovn.json', -5) === '.json'` is more safer.

src/wovnio/wovnphp/Utils.php
zeyuwu: I don't quite get this one. Can you explain it a bit more?

src/wovnio/html/HtmlConverter.php
```diff
@@ -273,7 +273,11 @@ private function buildHrefLang($lang_code)
             $url = $this->headers->removeLang($url, $this->store->defaultLang());
             $url = Url::addLangCode($url, $this->store, $lang_code, $this->headers);
         } else {
-            if ($lang_code !== $this->store->defaultLang()) {
+            $customDomainLangs = $this->store->getCustomDomainLangs();
+            $customDomainLang = $customDomainLangs ? $customDomainLangs->getCustomDomainLangByLang($lang_code) : false;
```
Soichiro725, line 277: I feel strange about that `$customDomainLang` can have a different type of variables like boolean and customDomainLang.

src/wovnio/wovnphp/Utils.php
Soichiro725: When client set like `$_SERVER['WOVN_CONFIG'] = '/path/to/wovn.json'`, WOVN.php will see wovn.json.
And we don't need to change here.
json user is miner now, we don't need to check wovn.json every time.

src/wovnio/html/HtmlConverter.php
Soichiro725: I prefer that.

test/unit/custom_domain/CustomDomainLangsTest.php
```diff
@@ -4,6 +4,7 @@
 require_once 'src/wovnio/wovnphp/custom_domain/CustomDomainLang.php';
 require_once 'src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php';
 
+use PHP_CodeSniffer\Tests\Standards\AllSniffs;
```
Soichiro725, line 7: Is this used?

zeyuwu: Good catch! 😬

src/wovnio/wovnphp/custom_domain/CustomDomainLangUrlHandler.php
```diff
@@ -7,13 +7,12 @@ public static function addCustomDomainLangToAbsoluteUrl($absoluteUrl, $targetlan
...
         $newLangCustomDomain = $customDomainLangs->getCustomDomainLangByLang($targetlang);
-        $changedUrl = self::changeToNewCustomDomainLang($absoluteUrl, $currentCustomDomain, $newLangCustomDomain);
-        return $changedUrl;
+        return self::changeToNewCustomDomainLang($absoluteUrl, $currentCustomDomain, $newLangCustomDomain);
     }
 
     public static function changeToNewCustomDomainLang($absoluteUrl, $currentCustomDomain, $newLangCustomDomain)
     {
-        if (!empty($currentCustomDomain) && !empty($newLangCustomDomain) && $currentCustomDomain->getLang() !== $newLangCustomDomain->getLang()) {
+        if (!empty($currentCustomDomain) && !empty($newLangCustomDomain)) {
```
bstewart00, line 15: what's the significance of removing this check? If both were the same language, it seems useless to continue.

test/unit/custom_domain/CustomDomainLangsTest.php
```diff
@@ -18,14 +17,35 @@ protected function setUp()
...
-        return $customDomainlang->getLang();
+        return $customDomainLang->getLang();
+    }
+
+    private function getHostAndPathWithoutTrailingSlash($customDomainLang)
+    {
+        return $customDomainLang->getHostAndPathWithoutTrailingSlash();
+    }
+
+    private function hashEquals($a, $b)
```
bstewart00, line 35: Is there no php built-in function that can do this?

src/wovnio/html/HtmlConverter.php
```diff
@@ -273,7 +273,14 @@ private function buildHrefLang($lang_code)
             $url = $this->headers->removeLang($url, $this->store->defaultLang());
             $url = Url::addLangCode($url, $this->store, $lang_code, $this->headers);
         } else {
-            if ($lang_code !== $this->store->defaultLang()) {
+            $customDomainLangs = $this->store->getCustomDomainLangs();
```
bstewart00, line 276: I think it would be better to split this logic into a separate else-if "pattern is custom domain" block.

src/wovnio/wovnphp/API.php
```diff
@@ -91,6 +93,18 @@ public static function translate($store, $headers, $original_content)
         }
     }
 
+    private static function getUriRepresentation($uri, $store, $lang)
+    {
+        $urlPatternName = $store->settings['url_pattern_name'];
+        $customDomainLangs = $store->getCustomDomainLangs();
```
bstewart00, line 99: move this inside the custom domain block, even if you need to nest another if afterwards. I'm not liking how this property is being checked for null. Can we ensure that `getCustomDomainLangs` is defined if the pattern is custom_domain?

src/wovnio/wovnphp/Url.php
```diff
@@ -251,7 +242,11 @@ public static function removeLangCode($uri, $lang_code, $store, $headers)
                 $customDomainLangs = $store->getCustomDomainLangs();
                 $default_lang = $settings['default_lang'];
                 $customDomainLangToRemove = $customDomainLangs->getCustomDomainLangByLang($lang_code);
-                $defaultCustomDomainLang = $customDomainLangs->getCustomDomainLangByLang($default_lang);
+                if ($customDomainLangToRemove->getSource()) {
```
bstewart00, line 245: I would perhaps suggest changing initialization a bit and ensuring that `->getSource()` is never null and points to the final resolved value.

When initializing the custom domain langs, use the source override if it has one, otherwise give it the default lang source.

src/wovnio/html/HtmlConverter.php
```diff
@@ -273,7 +273,14 @@ private function buildHrefLang($lang_code)
...
+            if ($customDomainLangs !== null) {
+                $customDomainLangHasSource = $this->store->getCustomDomainLangs()->hasSource($lang_code);
+            } else {
+                $customDomainLangHasSource = false;
+            }
+
+            if ($lang_code !== $this->store->defaultLang() || $customDomainLangHasSource) {
```
bstewart00, line 283: Remind me again why we need this extra check? So `removeLang` now always returns the physical URL (which is what `$this->headers->urlKeepTrailingSlash;` is). But when generating the hreflang, we need to convert this back into the virtual URL that is desired for the browser.

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
```diff
@@ -43,6 +65,44 @@ public function getCustomDomainLangByUrl($url)
...
+            $result[$lang->getHostAndPathWithoutTrailingSlash()] = $lang->getLang();
+        }
+        return $result;
+    }
+
+    /**
+     * Returns the computed (virtual) uri representation for a given physical location uri.
+     * Used when communicating with html-swapper.
+     *
+     * @param $uri string the current uri - pointing to physical location of current lang
```
bstewart00, line 81: rename variable to physicalUri to be even more clear.

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
```diff
@@ -43,6 +65,44 @@ public function getCustomDomainLangByUrl($url)
...
+        $currentLangDomainLang = $this->getSourceCustomDomainByLang($lang);
+        if ($currentLangDomainLang->getSource()) {
+            $defaultCustomDomainLang = $currentLangDomainLang->getSource();
+        } else {
+            $defaultCustomDomainLang = $this->getCustomDomainLangByLang($defaultLang);
+        }
+        return CustomDomainLangUrlHandler::changeToNewCustomDomainLang($uri, $currentLangDomainLang, $defaultCustomDomainLang);
+    }
+
+    public function hasSource($langCode)
```
bstewart00, line 97: Ideally I would like to remove this function.

test/unit/custom_domain/CustomDomainLangsTest.php
zeyuwu: Well there's the `array_diff_assoc` I can make use of. I guess I'll use that (but it still needs to be wrapped) since there's just no easy equality tests between associative arrays functioning as hash maps.

The original reason for why I hand coded this is I Googled "PHP hash map equals" and got the `hash_equals` built-in, and it's only available >5.6 so I decided against that. But ultimately it turns out it's not even doing what I expected, it literally compares the cryptographic hashes of two string?? D:

src/wovnio/wovnphp/custom_domain/CustomDomainLangUrlHandler.php
bstewart00: oh yeah, this is indeed a new scenario.

test/unit/custom_domain/CustomDomainLangsTest.php
bstewart00: in PHP parlance 'hash' is 'associative array' or 'multi-dimensional array' which is why you were getting results about string hashing. The hashes we deal with are only one level deep so we probably don't need a mega generic version.

zeyuwu: Yeah array_diff_assoc is clean enough, will use that.

src/wovnio/wovnphp/API.php
zeyuwu: We can enforce such a requirement but that would go against PHP's philosophy of "anything is better than nothing". Right now WOVN.php chugs along even if nothing is configured. If we choose to "halt' WOVN.php if a required setting is not fulfilled we'll have to make some big changes and make use of exceptions etc, which I don't think fits into the scope of this sprint.

src/wovnio/wovnphp/Utils.php
```diff
@@ -8,7 +8,14 @@ class Utils
     // will return the store and headers objects
     public static function getStoreAndHeaders(&$env)
     {
-        $file = isset($env['WOVN_CONFIG']) ? $env['WOVN_CONFIG'] : DIRNAME(__FILE__) . '/../../../../wovn.ini';
+        if (isset($_SERVER['WOVN_CONFIG'])) {
+            $file = $_SERVER['WOVN_CONFIG'] ? $_SERVER['WOVN_CONFIG'] : DIRNAME(__FILE__) . '/../../../../wovn.json';
```
Soichiro725, line 12: I guess this line should use `$env`.

zeyuwu: Yep. Just curious: why do we need to "rename" $_SERVER into $env here? It seems that $env here is purely an alias for $_SERVER anyways.

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
zeyuwu: We must test for if a custom domain have a source set, otherwise will just convert "no lang" url back to ones that has lang.

src/wovnio/html/HtmlConverter.php
zeyuwu: Yep you are correct in that reasoning.

src/wovnio/wovnphp/Utils.php
Soichiro725: Yes.
I guess there are two purpose.
- Change $_SERVER inside method. PHP can pass reference by adding `&` to the arguments.
- Make it easy to test, because test don't need to care about $_SERVER and just check arguments.

src/wovnio/wovnphp/Url.php
zeyuwu: I wish I've seen this comment earlier. Yep, doing that could save a lot of conditional checks.

zeyuwu: So like, we just have a fallback value for `source`? If you don't set one we set it to the `source` of default lang. And if default lang doesn't have a `source` then its `source` is itself.

src/wovnio/wovnphp/Utils.php
```diff
@@ -8,7 +8,14 @@ class Utils
     // will return the store and headers objects
     public static function getStoreAndHeaders(&$env)
     {
-        $file = isset($env['WOVN_CONFIG']) ? $env['WOVN_CONFIG'] : DIRNAME(__FILE__) . '/../../../../wovn.ini';
+        if (isset($env['WOVN_CONFIG'])) {
```
Soichiro725, line 11: If user set `/path/to/wovn.json` into `$_SERVER['WOVN_CONFIG']`, you don't need to change around here.
Just `$file = isset($env['WOVN_CONFIG']) ? $env['WOVN_CONFIG'] : DIRNAME(__FILE__) . '/../../../../wovn.ini';`  works.

zeyuwu: When I initially wrote this I thought of it as having two sources of settings, one from $env and one from $_SERVER. And after fixing $_SERVER to $env I forgot to remove the duplicate fragments. Sorry about that, not at my finest today.

src/wovnio/wovnphp/API.php
```diff
@@ -91,6 +93,22 @@ public static function translate($store, $headers, $original_content)
         }
     }
 
+    private static function getUriRepresentation($uri, $store, $lang)
+    {
+        $urlPatternName = $store->settings['url_pattern_name'];
+
+        if ($urlPatternName == 'custom_domain') {
+            $customDomainLangs = $store->getCustomDomainLangs();
```
bstewart00, line 101: When is this null? Ideally this would be validated during startup, but it seems we don't do any settings validation at all.

bstewart00: Not suggesting any changes though. It's just a bit annoying to need to check for null everywhere.

src/wovnio/wovnphp/Url.php
bstewart00: yep, kind of pre-compute the value.

src/wovnio/wovnphp/API.php
zeyuwu: I guess it's never null unless the constructor for CustomDomainLangs have some fatal error? And it's true that it's a big problem in WOVN.php that we don't do validations.

src/wovnio/html/HtmlConverter.php
```diff
@@ -272,10 +272,10 @@ private function buildHrefLang($lang_code)
         if ($this->store->hasDefaultLangAlias()) {
             $url = $this->headers->removeLang($url, $this->store->defaultLang());
             $url = Url::addLangCode($url, $this->store, $lang_code, $this->headers);
-        } else {
-            if ($lang_code !== $this->store->defaultLang()) {
-                $url = Url::addLangCode($url, $this->store, $lang_code, $this->headers);
-            }
+        } elseif ($lang_code !== $this->store->defaultLang()) {
+            $url = Url::addLangCode($url, $this->store, $lang_code, $this->headers);
+        } elseif ($this->store->getCustomDomainLangs() !== null) {
```
bstewart00, line 277: Since this no longer does `->getSource` stuff anymore, I think I have changed my mind on this and maybe we should combine the two ifs. However I don't really like using `$this->store->getCustomDomainLangs()` as a proxy for whether custom domain is enabled. It is possible for someone to have the settings in their config but still use path pattern etc. So I would prefer to just check the pattern name explicitly.

(the other usages of this for serializing the params etc is fine).

src/wovnio/wovnphp/API.php
bstewart00: It will be null if they use custom domain pattern but don't define the custom_domain_langs config. In which case they have bigger problems. In other backends we try to validate the settings and refuse to start.

src/wovnio/html/HtmlConverter.php
zeyuwu: Good catch.

src/wovnio/wovnphp/API.php
zeyuwu: Making PHP refuse to run is very difficult...

Soichiro725:
Can you add integration tests like test/integration/UrlCustomDomainPatternTest.php ?

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
```diff
@@ -43,9 +66,52 @@ public function getCustomDomainLangByUrl($url)
...
+        foreach ($customDomainLangsSettingsArray as $langUrl => $config) {
+            if (is_array($config)) {
+                $source = array_key_exists('source', $config) ? $config['source'] : $langUrl;
+                $lang = $config['lang'];
+            } else {
+                $lang = $config;
+                $source = $langUrl;
+            }
+
+            if ($lang === $defaultLang) {
```
bstewart00, line 111: This feels a bit backwards to me. What do you think about keying the hash by language instead?
```
        $this->customDomainLangsSetting = array(
            'foo.com/' => 'fr',
            'foo.com/path' => array('lang' => 'ja', 'source' => 'japan.foo.com/'),
            'foo.com/dir/path' => 'zh-CHS',
            'english.foo.com/' => array('lang' => 'en', 'source' => 'global.foo.com/')
        );

        $this->customDomainLangsSetting = array(
			'fr' => 'foo.com',
			'ja' => array('url' => 'foo.com/path', 'source' => 'japan.foo.com'),
			'zh-CHS' => 'foo.com/dir/path',
			'en' => array('url' => 'english.foo.com', 'source' => 'global.foo.com')
        );
```

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
```diff
@@ -43,9 +66,52 @@ public function getCustomDomainLangByUrl($url)
...
+    }
+
+    /**
+     * Returns the computed (virtual) uri representation for a given physical location uri.
+     * Used when communicating with html-swapper.
+     *
+     * @param $physicalUri string the current uri - pointing to physical location of current lang
+     * @param $lang string lang code of the current uri
+     * @param $defaultLang string lang code of the default (source) language
+     * @return string|string[]|null
```
bstewart00, line 85: Can the return value really be string or a string array? if so, this is not very ideal.

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
zeyuwu: I feel we do more lookup based on lang code so I guess that's good.

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
zeyuwu: Actually yes. If they caller pass in an array to $physicalUri the return value would be an array of strings. (not the intended use but it works that way.)

zeyuwu: Could add a type check for input params but I don't think it's worth it. We would be basically fighting PHP itself.

bstewart00: i'm confused still. Isn't `$physicalUri` always a string?  what's an example of an array return value?

zeyuwu: Sorry I didn't phrase this clearly. Inside the usage of WOVN.php $physicalUri is never an array, it is always a string. But because inside `changeToNewCustomDomainLang`, which `computeSourceVirtualUrl` calls, there's a call to `preg_replace` that's capable of accepting and returning an array. And since PHP doesn't do type checking it makes the `computeSourceVirtualUrl` capable of returning an array as well. 

I'll just update the doc string to remove the string[] to make things clearer.

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
```diff
@@ -6,14 +6,37 @@
...
 
             // Disable notice error by adding @, when path is not defined
-            array_push($this->customDomainLangs, new CustomDomainLang($parsedUrl['host'], @$parsedUrl['path'], $lang));
+            array_push($this->customDomainLangs, new CustomDomainLang($parsedUrl['host'], @$parsedUrl['path'], $config['lang'], $source));
+        }
+    }
+
+    public function getSourceCustomDomainByLang($langCode)
+    {
+        $results = array_filter($this->customDomainLangs, function ($customDomain) use ($langCode) {
```
bstewart00, line 31: This function looks very similar to `getCustomDomainLangByLang`. I suspect you could be using it internally.

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
bstewart00: Yeah, if preg_replace only ever returns string in the situations we use it in, just ignore the array case.

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
bstewart00: note that I think the json format should also follow this if we do it.

zeyuwu: If we do that there's the drawback that our json format and ini format no longer matches each other.

But if we decide to do that, can I suggest to group other info together under the language key too? Like I proposed earlier:

```
{
    "langauges": {
        "en": {
            "is_source": false,
            "physical_url": null,
            "url": {
                "domain": "global.example.com",
                "path": null,
                "query": {
                    "key": "lang",
                    "value": "en"
                }
            }
        },
        "fr": {
            "is_source": false,
            "physical_url": null,
            "url": {
                "domain": "global.example.com",
                "path": null,
                "query": {
                    "key": "lang",
                    "value": "fr"
                }
            }
        },
        "ja": {
            "is_source": true,
            "physical_url": {
                "domain": "example.co.jp",
                "path": null,
                "query": null
            },
            "url": {
                "domain": "example.ne.jp",
                "path": null,
                "query": null
            }
        },
        "cn": {
            "is_source": true,
            "physical_url": null,
            "url": {
                "domain": "example.ne.jp",
                "path": "/cn/",
                "query": null
            }
        }
    }
    
}
```

bstewart00: Custom domain isn't supported at all in the ini format I thought (or at least we shouldn't support it)

zeyuwu: We do support "custom domain classic" in the old ini format, there are special handling in the code to make that happen.

bstewart00: I think that kind of format might be going too far right now. I would save it for if we decide to overhaul the url pattern system.

bstewart00: Yeah, I would remove support for custom domain classic in this change 😉

zeyuwu: I mean why don't we just keep quiet about this feature and just push JSON to anyone who wants custom domain, because... we don't exactly have a good way to make PHP stop service? Like, we don't really have a way to "remove" support unless we check specifically when reading from .ini for that keyword and raise a fatal error.

bstewart00: Sorry, I don't really mean remove support in the code I guess. It is now undefined behavior to add a custom_domain_langs property to your ini file. Done? Everyone who is going to use this feature will be instructed to use the json format 😄

zeyuwu: I'll add that into the documentation for WOVN.php.

test/helpers/TestUtils.php
```diff
@@ -135,4 +143,11 @@ public static function enableRewritePathPattern($htaccessFilePath, $langIdentifi
             file_put_contents($htaccessFilePath, $replacedHtaccess);
         }
     }
+
+    public static function enableWovnJson($htaccessFilePath)
+    {
+        $htaccess = file_get_contents($htaccessFilePath);
+        $replacedHtaccess = $htaccess . "\nSetEnv WOVN_CONFIG";
+        file_put_contents($htaccessFilePath, $replacedHtaccess);
+    }
```
Soichiro725, line 152: SetEnv in .htaccess means that set variable into `$_SERVER['WOVN_CONFIG']`;
So, this method should be like the following.
```
public static function enableWovnJson($htaccessFilePath, $configFilePath)
    {
        $htaccess = file_get_contents($htaccessFilePath);
        $replacedHtaccess = $htaccess . "\nSetEnv WOVN_CONFIG " . $configFilePath;
        error_log($replacedHtaccess);
        file_put_contents($htaccessFilePath, $replacedHtaccess);
    }
```

test/integration/UrlCustomDomainPatternTest.php
```diff
@@ -170,6 +170,49 @@ public function testCustomDomainPatternSubDir()
         $this->assertEquals('<html><head></head><body>html-swapper-mock</body></html>', TestUtils::fetchURL('http://zh-hant-hk.com/zh/sub/index.html')->body);
     }
 
+    public function testCustomDomainPatternWithDomainSource()
+    {
+        copy("{$this->sourceDir}/wovn_index_sample.php", "{$this->docRoot}/wovn_index.php");
+        mkdir("{$this->docRoot}/ja");
+        TestUtils::writeFile("{$this->docRoot}/ja/index.html", '<html><head></head><body>test</body></html>');
+        TestUtils::enableWovnJson("{$this->docRoot}/.htaccess");
```
Soichiro725, line 178: I think `setWovnConfig` is better for this method.
```
TestUtils::enableWovnJson("{$this->docRoot}/.htaccess", "{$this->docRoot}/wovn.json");
```

bstewart00 (approved):
nice work, we got there in the end 😄

test/helpers/TestUtils.php
zeyuwu: Yeah, I see what you mean. For this one I took a shortcut by simply setting the env variable (leaving it blank). And when `_SERVER['WOVN_CONFIG']` is set but is blank, WOVN.php will try to load `wovn.ini` from the default location of `{document_root}/wovn.ini`. It's a bit hacky. 😅 Will update it to the proper way.

Soichiro725 (approved):

Soichiro725:
Great job!!

PR #135, opened 2020-10-08
set originalPath based on REQUEST_URI. Fix custom domain tests.

bstewart00:
### Comments
I didn't like how `originalPath` was computed using REDIRECT_URL, but the rest of our pathname handling uses REQUEST_URI. I think we can combine both of these together.

Improved custom domain test cases by setting up HTTP_HOST/REQUEST_URI env variables for each test case.

### DIFF
--- a/test/unit/HeadersTest.php
+++ b/test/unit/HeadersTest.php
@@ -151,58 +151,65 @@ class HeadersTest extends \PHPUnit_Framework_TestCase
             'custom_domain_langs' => $custom_domain_langs
         );
         $testCases = array(
-            // beforeRemoveUrl, afterRemoveUrl, removeLang
-            array('http://testsite.com', 'http://testsite.com', 'en'),
-            array('http://en-us.testsite.com', 'http://testsite.com', 'en-US'),
-            array('http://testsite.com/ja', 'http://testsite.com', 'ja'),
-            array('http://testsite.com/zh/chs', 'http://testsite.com', 'zh-CHS'),
-            array('http://zh-hant-hk.com/zh', 'http://testsite.com', 'zh-Hant-HK'),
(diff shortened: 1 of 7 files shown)


test/unit/HeadersTest.php
```diff
@@ -193,8 +193,7 @@ public function testRemoveLangWithCustomDomainPattern()
             array('/zh/chs/', '/', 'zh-CHS'),
         );
         $env = array(
-            'HTTP_HOST' => 'testsite.com',
-            'REQUEST_URI' => '/zh/'
```
Soichiro725, line 197: I found my mistake about test.
I wanted to test, when accessed by zh lang.
So, I should do `array('zh-hant-hk.com', 'testsite.com', 'zh-Hant-HK')` not `array('zh-hant-hk.com', 'zh-hant-hk.com', 'zh-Hant-HK')` in testRemoveLangWithCustomDomainPattern.
And in this case, `'REQUEST_URI' => '/zh/'` is needed.

bstewart00: I think the only way we can be sure about these tests is to include REQUEST_URI, HTTP_HOST in the test cases.
```
// only path
array('/zh/', '/zh/', 'zh-Hant-HK'),
```
This one for example. `/zh/` should become `/` if HTTP_HOST was `zh-hant-hk.com`, otherwise `/zh/`. I think we should be testing this.

Soichiro725: Yes. It is better to set some $_SERVER env per test.
I felt annoying, so I set one REQUEST_URI :stuck_out_tongue_closed_eyes:

Soichiro725 (approved):

zeyuwu (approved):
I've always felt that the complicated header.php initialization _must_ have been because of some consideration for weird proxy setups. But I guess we could do away with those now.

PR #139, opened 2020-10-21
Feature/server cookie lang

zeyuwu:
https://wovnio.atlassian.net/browse/ST-510

Adds support for WOVN.php to respect cookie language settings.

The backend will send a 302 Redirect if a language cookie is set, only if when the visited URL is in the default language. In this case, the visitor will be redirected to a URL of the cookie language.

NEW CONFIG:
use_cookie_lang (boolean).


src/wovnio/wovnphp/CookieLang.php
```diff
@@ -0,0 +1,61 @@
...
+            return $_COOKIE[CookieLang::COOKIE_LANG_NAME];
+        }
+        return null;
+    }
+
+    public function shouldRedirect() {
+        return self::getCookieLang() && ($this->header->lang() != self::getCookieLang());
+    }
+
+    public function computeRedirectUrl()
```
bstewart00, line 42: Feel like this should not be the responsibility of CookieLang.

bstewart00: in this case Headers.responseOut probably _is_ the correct place for dealing with redirects.

zeyuwu: Indeed it shouldn't. The logic were taken directly out of HtmlConverter. I tried to make it public from HtmlConverter but its requiring way too many constructor arguments to be practical in CookieLang. Perhaps it's better if I move it to Url?

bstewart00: This is mostly delegating to headers methods here, which is why I think headers is the best place for it. Url is supposed to only be about manipulating lang codes on URLs.

src/wovnio/wovnphp/CookieLang.php
```diff
@@ -0,0 +1,36 @@
...
+    public function getCookieLang()
+    {
+        $cookies = $this->header->getCookies();
+        if (array_key_exists(CookieLang::COOKIE_LANG_NAME, $cookies)) {
+            return $cookies[CookieLang::COOKIE_LANG_NAME];
+        }
+        return null;
+    }
+
+    public function shouldRedirect()
```
bstewart00, line 29: Still do not like this being here. Make CookieLang simply be about telling us what the value is of the cookie. How we actually use this cookie and whether the feature is enabled is best left to the code doing the redirection.

src/wovnio/wovnphp/Headers.php
```diff
@@ -315,4 +337,31 @@ public function getDocumentURI()
...
+            $originalURL = $originalURL . '?' . $this->query;
+        }
+        return $originalURL;
+    }
+
+    public function computeRedirectUrl()
+    {
+        $cookieLangCode = $this->cookieLang->getCookieLang();
+        $url = $this->urlKeepTrailingSlash;
+        if ($this->store->hasDefaultLangAlias()) {
```
bstewart00, line 354: This looks exactly the same `buildHrefLang`

```
    private function buildHrefLang($lang_code)
    {
        $url = $this->headers->urlKeepTrailingSlash;

        if ($this->store->hasDefaultLangAlias()) {
            $url = $this->headers->removeLang($url, $this->store->defaultLang());
            $url = Url::addLangCode($url, $this->store, $lang_code, $this->headers);
        } elseif ($lang_code !== $this->store->defaultLang() || $this->store->settings['url_pattern_name'] === 'custom_domain') {
            $url = Url::addLangCode($url, $this->store, $lang_code, $this->headers);
        }
        return htmlentities($url);
    }
```

Then we need to think of a name for this and move this function somewhere.

bstewart00: I would leave out the first line and make an `addLangCodeToUrl($url)` function.

bstewart00: Actually, it seems like Url.php already has handling for the defaultLang alias. So I wonder if this special case is even needed anymore. Just call `Url.addLangCode`.

bstewart00: Also not sure if you need to apply `htmlentities` for your redirect URL. The hreflang does it because it is being written to a HTML attribute.

zeyuwu: Yep, I'll go for the shortened version with Url::addLangCode.

README.en.md
```diff
@@ -486,6 +486,13 @@ This parameter is required when you set `enable_wovn_diagnostics` to `true`. Thi
 wovn_diagnostics_password = wovn_diagnostics_password
 ```
 
+#### `use_cookie_lang`
+When set to true, WOVN.php will attempt to redirect a request to the default language to the language set in the `wovn_selected_lang` cookie. No redirect will happen if no such cookie is set.
+
+```
+use_cookie_lang = ture
```
bstewart00, line 493:
```suggestion
use_cookie_lang = true
```

src/wovn_interceptor.php
```diff
@@ -43,7 +45,10 @@
     }
     // use the callback of ob_start to modify the content and return
     ob_start(function ($buffer) use ($headers, $store, $diagnostics, $benchmarkStart) {
-
+        if ($headers->shouldRedirect()) {
+            // this carries an implied HTTP 302
+            header("Location: " . $headers->computeRedirectUrl());
```
bstewart00, line 50: https://stackoverflow.com/questions/3553698/php-should-i-call-exit-after-calling-location-header my assumption is we should exit() after such a redirect.

src/wovnio/wovnphp/Headers.php
```diff
@@ -252,22 +256,39 @@ private function removeLangFromQuery()
...
+            if (array_key_exists($locationHeader, $responseHeaders)) {
+                $redirectLocation = $responseHeaders[$locationHeader];
             }
         }
+
+        if ($redirectLocation) {
+            $newLocation = Url::addLangCode($redirectLocation, $this->store, $urlLanguage, $this);
+        }
+
+        if ($this->shouldRedirect()) {
```
bstewart00, line 282: i'm confused why this code is here when I just saw in `wovn_interceptor` the same code checking this and sending the redirect?

src/wovn_interceptor.php
zeyuwu: I also thought there has to be some kind of termination there. I tried `fastcgi_finish_request` but it's not working. That's kinda also the reason why I did the redirect in the header too, to fix the behavior.

src/wovnio/wovnphp/Headers.php
```diff
@@ -19,18 +19,23 @@ class Headers
...
+     * @param array $cookies &$cookies Contains the _COOKIE env variable
      * @return void
      */
-    public function __construct(&$env, &$store)
+    public function __construct(&$env, &$store, $cookies)
     {
         $this->env =& $env;
         $this->store =& $store;
+        $this->cookies = $cookies;
+        $this->cookieLang = new CookieLang($this);
```
Soichiro725, line 38: I think WOVN.php doesn't need to keep all cookies.
So, my suggestion is remove `$this->cookies = $cookies;`.
And set `$cookies` to `CookieLang` and only keep `$this->cookieLang` .

test/unit/CookieLangTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->assertEquals(false, $headers->shouldRedirect());
+    }
+
+    public function testShouldRedirectDefaultLangQueryPattern()
+    {
+        list($store, $headers) = $this->getHeaderStoreQueryPattern('ja', 'en');
+        $this->assertEquals(true, $headers->shouldRedirect());
+    }
+
+    public function testShouldNotRedirectNonSourceToNonSourcePathPattern()
```
bstewart00, line 75: "TargetToTarget" would be more commonly used terms than NonSource

src/wovnio/wovnphp/Headers.php
zeyuwu: The reason I kept a separate copy of cookies (which came from the superglobal $_COOKIE, is to facilitate testing, so that we can easily mock a cookie value. The same way $env is used.

test/helpers/TestUtils.php
```diff
@@ -30,21 +30,27 @@ public static function addHost($host)
         file_put_contents('/etc/hosts', $hostFile);
     }
 
-    public static function fetchURL($url, $timeout = 3)
+    public static function fetchURL($url, $timeout = 3, $langCookie = null)
```
bstewart00, line 33: I'm finding these tests really hard to read with `$langCookie` being a string parameter. I had no idea `fetchURL($url, 3, 'en')` sets the cookie language. Can't we pass in an array of cookies in the test ? `array('wovn_selected_lang' => 'en')` that is very clear.

zeyuwu: Good point.

test/integration/CookieLangTest.php
```diff
@@ -0,0 +1,98 @@
...
+        mkdir("{$this->docRoot}/v0");
+        copy("{$this->sourceDir}/test/fixtures/integration/v0/translation", "{$this->docRoot}/v0/translation");
+    }
+
+    protected function tearDown()
+    {
+        TestUtils::cleanUpDirectory($this->docRoot);
+    }
+
+    public function testShould302RedirectDefaultLangQueryPattern()
```
bstewart00, line 37: I generally don't like test names that don't describe the actual scenario and force you to read the test body. Unfortunately this is the current pattern of the existing tests in wovn.php

This is testing that when browsing to a default lang url with a wovn cookie, a redirection is sent.
`testRequestToDefaultLang__WithCookie__ShouldRedirect` something like this would be my choice here.

Similarly the ShouldNotRedirect tests are more like
`testRequestToTargetLang__WithCookie__ShouldNotRedirect`

I'm having a hard time figuring out what `testShouldNotRedirectDefaultLangToOtherLangPathPattern` tests by the name.
It's testing that a request to a target language with default lang cookie doesn't redirect to the cookie lang? Not obvious at all from the name.

test/helpers/TestUtils.php
bstewart00: Might also be a good idea to make `$timeout = null` the default, and do a `= 3` inside the method. Just so we could pass in null instead of an arbitrary 3 inside of tests if we don't care.

src/wovn_interceptor.php
bstewart00: I guess even the redirect we already do did not do this. So maybe we don't need to.

src/wovnio/wovnphp/Headers.php
bstewart00: I would also argue Headers does not need a  copy of $_COOKIE at all. Headers only has it to give to CookieLang. Just give it to CookieLang. Which would also solve your dependency issue.

test/integration/CookieLangTest.php
zeyuwu: New names adopts a requested lang + cookie status + url pattern + expected behavior format. Should be friendlier now.

src/wovnio/wovnphp/Headers.php
bstewart00: Looks good. I think it doesn't use Headers anymore though thus can be simplified further like I mentioned.

zeyuwu: Good catch. I must be blind.

src/wovnio/wovnphp/Headers.php
```diff
@@ -315,4 +328,34 @@ public function getDocumentURI()
...
+
+    public function computeRedirectUrl()
+    {
+        $cookieLangCode = $this->cookieLang->getCookieLang();
+        $url = $this->urlKeepTrailingSlash;
+        $url = Url::addLangCode($url, $this->store, $cookieLangCode, $this);
+        return $url;
+    }
+
+    public function getCookies()
```
Soichiro725, line 349: Now, `getCookies` is not used, so this can be deleted.

test/unit/CookieLangTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $cookies = array(
+            CookieLang::COOKIE_LANG_NAME => $cookieLang
+        );
+        return StoreAndHeadersFactory::fromFixture('default', $settings, $env, $cookies);
+    }
+
+    public function testRequestToTargetLangWithTargetCookieQueryPatternShouldNotRedirect()
+    {
+        list($store, $headers) = $this->getHeaderStoreQueryPattern('ja', 'fr');
+        $this->assertEquals(false, $headers->shouldRedirect());
```
bstewart00, line 54: These tests really ought to be in HeadersTest

src/wovnio/wovnphp/Headers.php
```diff
@@ -315,4 +328,29 @@ public function getDocumentURI()
 
         return $this->removeLang($url, $this->requestLang());
     }
+
+    public function constructOriginalURL()
```
bstewart00, line 332: looks unused

src/wovnio/wovnphp/Headers.php
```diff
@@ -315,4 +328,29 @@ public function getDocumentURI()
...
+        $url = Url::addLangCode($url, $this->store, $cookieLangCode, $this);
+        return $url;
+    }
+
+    public function shouldRedirect()
+    {
+        if (!$this->store->settings['use_cookie_lang']) {
+            return false;
+        }
+        return $this->cookieLang->getCookieLang() && ($this->requestLang() != $this->cookieLang->getCookieLang()) && $this->requestLang() === $this->store->defaultLang();
```
bstewart00, line 354: extract `$cookieLang, $requestLang` variables instead of calling these twice. I think doing the requestLang is default check first might make more sense.
```

return $requestLang === $this->store->defaultLang()
	&& $cookieLang
	&& $this->requestLang() != $cookieLang
```

src/wovnio/wovnphp/Headers.php
```diff
@@ -19,18 +19,22 @@ class Headers
...
      * @param Store &$store The store containing user settings
+     * @param array $cookies &$cookies Contains the _COOKIE env variable
      * @return void
      */
-    public function __construct(&$env, &$store)
+    public function __construct(&$env, &$store, $cookies)
     {
         $this->env =& $env;
         $this->store =& $store;
+        $this->cookieLang = new CookieLang($cookies);
```
bstewart00, line 37: `private $cookies` looks unused. You should also just replace the `$cookies` parameter with CookieLang.

src/wovnio/wovnphp/Headers.php
bstewart00: actually I guess the $cookieLang check is the simplest so how you had it is fine. Take this as a formatting suggestion though.

src/wovnio/wovnphp/Headers.php
zeyuwu: Private $cookies indeed is unused, apologies for all the unused stuff. My IDE's intelisense/code highlighting just refused to work properly somehow.

But I think it should be called $cookies, since it's the $_COOKIE super global we are effectively passing into the constructor there.

bstewart00: I think that's fine. The CookieLang class is more of a wrapper anyway and could be called Cookies even 😉

zeyuwu: Yeah, after moving the bulk of logic into headers I agree.

bstewart00 (approved):

Soichiro725 (approved):

PR #89, opened 2019-07-03
Support HTTP Header X-Forwarded-Request-Uri for when proxy is used

torkelbe:
### Purpose/goal of PR (Issue #)
When WOVN.php serves a request behind a proxy and the request URI gets modified in the process, it will need to read the original request URI from a HTTP Header. The header name for original URI is `X-Forwarded-Request-Uri`, and it must be manually set by the user. See the README for an example of how set it.

The original client requested URI is needed in order to correctly identify the requested page, and is used for setting hreflang tags in the HTML content and for locating the correct translation data.

WOVN.php will only look at the `X-Forwarded-Request-Uri` header if `use_proxy = 1` is set in wovn.ini.

PR #140, opened 2020-10-21
Refactor environment variable access

bstewart00:
### Comments
Goal is to simplify the proxy scenario where `HTTP_X_FORWARDED_REQUEST_URI` is used instead of `REQUEST_URI` etc.

### DIFF
--- a/src/wovn_interceptor.php
+++ b/src/wovn_interceptor.php
@@ -1,6 +1,7 @@
 <?php
 require_once 'version.php';
 require_once 'wovnio/wovnphp/Headers.php';
+require_once 'wovnio/wovnphp/Environment.php';
 require_once 'wovnio/wovnphp/Lang.php';
 require_once 'wovnio/wovnphp/Logger.php';
 require_once 'wovnio/wovnphp/Store.php';
--- a/src/wovnio/wovnphp/Environment.php
+++ b/src/wovnio/wovnphp/Environment.php
@@ -1,0 +1,129 @@
+<?php
+namespace Wovnio\Wovnphp;
+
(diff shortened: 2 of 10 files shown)


bstewart00:
The one change in behavior i'm slightly unsure about is in `removeLangFromPath`, we used to remove lang from both `REQUEST_URI` and `X_FORWARDED_REQUEST_URI` if they are set. Now we only set either one or the other.

i've been reading https://github.com/WOVNio/equalizer/issues/16965 and https://github.com/WOVNio/WOVN.php/pull/89 , and my understanding of how the proxy scenario works is:
1. First server gets the request, htaccess rewrite happens before proxy (thus why X_FORWARDED_REQUEST_URI was needed)
2. Second server receives request, REQUEST_URI already has the lang removed, and therefore doesn't need further modification by wovn.php. Language is detected using X_FORWARDED_REQUEST_URI.

@a-bougouin do you happen to remember how these proxy variables work? Good or bad idea?

a-bougouin:
@bstewart00 I am not sure that we can count on the reverse proxy to always remove the language information from the location (`REQUEST_URI` on second server). The issue you mention was a specific scenario for which what you said would work, but how about other reverse proxies that would for instance just change the domain and leave the path unchanged?

bstewart00:
Good point, we should probably prefer to be cautious here. In which case I think this strategy of hiding environment variables won't work. With this design, probably the only thing we could do is set both `REQUEST_URI` and the forwarded version to the same value, rather than applying `removeLang` to them individually which seems safer.

I will probably put this on hold and think on this more.

PR #144, opened 2020-11-03
Fix wovn-ignore mark

Soichiro725:
`$node->getAttribute('wovn-ignore')` is falsy, even if the attribute exists.
I added `hasAttribute` method to check having `wovn-ignore` or not.

```
<span wovn-ignore>xxx</span> -> marked
<span wovn-ignore="true">xxx</span> -> marked
<span wovn-ignore="">xxx</span> -> not marked (should mark)
```

### DIFF
--- a/src/wovnio/modified_vendor/SimpleHtmlDomNode.php
+++ b/src/wovnio/modified_vendor/SimpleHtmlDomNode.php
@@ -405,7 +405,7 @@ class SimpleHtmlDomNode {
     function __isset($name)
     {
         $val = $this->get_attr_val($this->attribute, $name);
-        return isset($val) && $val !== false ;
+        return $val !== false;
     }
 
   /*
--- a/src/wovnio/modified_vendor/SimpleHtmlDomNode.php
+++ b/src/wovnio/modified_vendor/SimpleHtmlDomNode.php
@@ -280,7 +280,7 @@ class SimpleHtmlDomNode {
     {
         $val = $this->get_attr_val($this->attribute, $name);
 
-        if ($val) {
+        if (isset($val)) {
             return $this->convert_text($val);
         }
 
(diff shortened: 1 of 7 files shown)


src/wovnio/modified_vendor/SimpleHtmlDomNode.php
```diff
@@ -404,15 +404,8 @@ function replaceAttributeValue($name, $value) {
...
-        {
-            case 'outertext': return true;
-            case 'innertext': return true;
-            case 'plaintext': return true;
-        }
-        //no value attr: nowrap, checked selected...
-        // return (array_key_exists($name, $this->attr)) ? true : isset($this->attr[$name]);
-        return false;
+        $val = $this->get_attr_val($this->attribute, $name);
+        return isset($val) && $val !== false ;
```
zeyuwu, line 408: Should false be considered not set?

Soichiro725: `isset(false)` is `true`.

Soichiro725: When attribute exists, $val will be value of attribute or `true`.
When attribute doesn't exist, $val will be `false`.

Soichiro725: Ah! I can replace by `return $val !== false ;`

zeyuwu: I see. I'm just a little confused about it because in JavaScript the `hasAtrribute()` function returns true even if the value of that attribute is false. I guess the use case in WOVN.php is slightly different in that we want to test if an attribute is explicitly defined as "valid".

image

Soichiro725: I think hasAttribute in WOVN.php will behave as same way as JavaScript.
You can see test cases in `testHasAttribute`.

zeyuwu: Ah I see. get_attr_val() will return a string "true" or "false". So it indeed behaves the same as JavaScript.

zeyuwu (approved):

iwai-wovn (approved):

bstewart00 (approved):

PR #147, opened 2020-11-30
Add insert_hreflangs feature

Soichiro725:
https://wovnio.atlassian.net/browse/ST-572

Don't touch hreflang, if `insert_hreflangs` is false.

WOVN.php has two logic to insert hreflang.
I removed it from `convertToAppropriateBodyForApi` and changed to only use `insertSnippetAndHreflangTags`.

### Comments
html-swapper part:
https://github.com/WOVNio/html-swapper/pull/541

### DIFF
--- a/test/unit/html/HtmlConverterTest.php
+++ b/test/unit/html/HtmlConverterTest.php
@@ -660,7 +660,7 @@ class HtmlConverterTest extends \PHPUnit_Framework_TestCase
         );
         list($store, $headers) = StoreAndHeadersFactory::fromFixture('default', $settings);
         $converter = new HtmlConverter(null, $store->settings['project_token'], $store, $headers);
-        list($translated_html, $marker) = $this->executeConvert($converter, $html, 'UTF-8', '_removeCustomIgnoreClass');
+        list($translated_html, $marker) = $this->executeConvert($converter, $html, 'UTF-8', 'removeCustomIgnoreClass');
         $keys = $marker->keys();
 
         $this->assertEquals(1, count($keys));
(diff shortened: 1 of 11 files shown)


test/unit/html/HtmlConverterTest.php
```diff
@@ -9,103 +9,163 @@
...
-
-        $actual_html_text = $marker->revert($translated_html);
-        $doc = new \DOMDocument("1.0", "ISO-8859-15");
-        $doc->loadHTML(mb_convert_encoding($actual_html_text, 'HTML-ENTITIES', "utf-8"));
-        $actual_html = $doc->saveHTML();
-
-        $this->assertEquals($expected_html, $actual_html);
+    public function testInsertSnippetAndHreflangTags()
+    {
+        $html_cases = array(
```
bstewart00, line 39: I think if you just want to re-use the code in the test, you can make a helper method for the common code. With this style of test you're basically making a "test framework inside of a test framework". It's already starting to get a bit complicated here so I would be careful.

bstewart00 (approved):

test/unit/html/HtmlConverterTest.php
Soichiro725: I want to do a parameterized test.
I also write as DRY.
When I modify code in WOVN.php, it is really hard.
Because there are many similar test cases, but sometimes there are no needed test cases.
I think a few function is easy to modify and compare test cases.

a-bougouin (approved):

Soichiro725:
I updated README.md.
Please fix me, if English is strange.

PR #148, opened 2020-12-02
Add debug request parameters

aw-leigh:
https://wovnio.atlassian.net/browse/ST-589

### Comments
Adds support for `wovnDisable` URL parameter which if enabled will cause the library to not swap the request.

Also adds a Debug Mode setting that enables two URL parameters:
`wovnCacheDisable`: inserts a timestamp into the API call url to circumvent Fastly caching
`wovnDebugMode`: adds {“debug_mode”: “true”} to API call request payload, and add a timestamp to the API call to circumvent Fastly cache.

Usage:
All parameters can enabled by adding to the URL either without a value (e.g. `?lang=ja&wovnCacheDisable`) or with any value other than the string `'false'` (e.g. `?lang=ja&wovnDebugMode=true`).

### DIFF
--- a/src/wovn_interceptor.php
+++ b/src/wovn_interceptor.php
@@ -32,7 +32,7 @@ if (!$store->isValid()) {
     return false;
 }
 
-$queryString = parse_url($_SERVER['REQUEST_URI'])['query'];
+$queryString = $_SERVER['REQUEST_URI'];
 $debugModeEnable = $store->settings['debug_mode'];
 $requestOptions = new RequestOptions($queryString, $debugModeEnable);
 
--- a/src/wovn_interceptor.php
+++ b/src/wovn_interceptor.php
@@ -59,7 +59,7 @@ if (!Utils::isIgnoredPath($uri, $store)) {
         $diagnostics = new Diagnostics($store);
     }
     // use the callback of ob_start to modify the content and return
-    ob_start(function ($buffer) use ($headers, $store, $diagnostics, $benchmarkStart) {
+    ob_start(function ($buffer) use ($headers, $store, $diagnostics, $benchmarkStart, $requestOptions) {
         if ($headers->shouldRedirect()) {
             // this carries an implied HTTP 302
             header("Location: " . $headers->computeRedirectUrl());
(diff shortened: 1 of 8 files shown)


src/wovn_interceptor.php
```diff
@@ -20,10 +21,15 @@
...
 use Wovnio\Wovnphp\API;
 use Wovnio\Wovnphp\Diagnostics;
+use Wovnio\Wovnphp\RequestOptions;
 use \Wovnio\Wovnphp\CookieLang;
 
 // GET STORE AND HEADERS
 list($store, $headers) = Utils::getStoreAndHeaders($_SERVER, $_COOKIE);
+// RequestOptions here?
+
+$queryString = parse_url($headers->getEnv()['REQUEST_URI'])['query'];
```
bstewart00, line 31: yes, request options here but you can use $_SERVER directly instead of headers->getEnv

bstewart00: well, I guess we could make $headers->query public as well, regrettably

aw-leigh:
Since API::translate is just a static function call here (as opposed to a method in an API object like in Wovnjava), I added a `$cache_disable_mode` parameter. Is this alright or is there a better way I'm overlooking? If it's alright then I'll have to edit the calls in APITest, but I wanted to double check before doing so. Or I guess I could make `$cache_disable_mode` false by default, but I'm told that default parameters are to be avoided if possible.

bstewart00:
yes, using a parameter instead is the usual method

src/wovnio/wovnphp/RequestOptions.php
```diff
@@ -0,0 +1,57 @@
...
+
+
+    public function __construct($queryString, $debugModeEnable)
+    {
+        $this->disableMode = false;
+        $this->cacheDisableMode = false;
+        $this->debugMode = false;
+
+        if ($queryString !== null) {
+            $this->disableMode = strpos($queryString, 'wovnDisable') !== false;
```
bstewart00, line 35: note that this is a bit naive. We would like to allow `wovnDisable=false/true` ideally. Which is why making `$headers->query` public (+ https://www.php.net/manual/en/function.parse-str.php perhaps to turn it into an array) might be a better approach.

aw-leigh: I can do that, though I'm not sure when we'd ever want to explicitly set wovnDisable to false.

bstewart00: It's handy when you're debugging to switch between on/off quickly

bstewart00: Didn't realize wovnjava also behaves like this though.

aw-leigh: Yeah that's why I did it that way 🙃

zeyuwu (approved):

bstewart00 (approved):

PR #151, opened 2020-12-14
Feature/insert html lang attribute

Soichiro725:
https://wovnio.atlassian.net/browse/ST-599

Add lang attribute to html tag.


zeyuwu (approved):
Looks good.

aw-leigh (approved):

src/wovnio/html/HtmlConverter.php
```diff
@@ -55,6 +55,18 @@ public function insertSnippetAndHreflangTags($html, $add_fallback_mark)
         return $converted_html;
     }
 
+    public function insertHtmlLangAttribute($html, $lang_code)
+    {
+        if (preg_match('/<html\s?.*?>/', $html, $matches)) {
+            $html_open_tag = $matches[0];
+            $replacement = $matches[0];
+            $replacement = preg_replace('/lang=["\']?[a-zA-Z-]*["\']?/', '', $replacement); // remove exsiting lang attribute
```
bstewart00, line 63: if this is only getting called if we are not making an API call, should we be keeping the existing lang tag instead? We aren't actually changing the language in this case.

test/unit/html/HtmlConverterTest.php
```diff
@@ -238,6 +238,55 @@ public function testInsertSnippetAndHreflangTagsWithFallback()
...
+                '<html><head></head><body><a>hello</a></body></html>',
+                '<html lang="en"><head></head><body><a>hello</a></body></html>',
+            ),
+            array(
+                'lang attribute is exist',
+                '<html lang="ja"><head></head><body><a>hello</a></body></html>',
+                '<html lang="en" ><head></head><body><a>hello</a></body></html>',
+            ),
+            array(
+                'lang attribute is exist with single quatation',
```
bstewart00, line 255:
```suggestion
                'lang attribute exists with single quotes',
```

src/wovnio/html/HtmlConverter.php
Soichiro725: Now, I think library should always insert default lang attr.
Because, if HTML translated properly, html-swapper changed lang attr.

bstewart00: I'm more thinking if the request is for a default lang and not using html swapper, and they have `lang=fr` in their HTML already. Should we keep it or not?

Soichiro725: I guess that case is kind of client's mistake for setting WOVN.php.
But I agree to not force our lang setting.
I will change to not overwrite lang attr.

bstewart00 (approved):

src/wovnio/wovnphp/API.php
```diff
@@ -45,6 +47,7 @@ public static function translate($store, $headers, $original_content, $request_o
             $converted_html = $converter->convertToAppropriateBodyForApi($converted_html);
         }
         $converted_html = $converter->insertSnippetAndHreflangTags($converted_html, true);
+        $converted_html = $converter->insertHtmlLangAttribute($converted_html, $default_lang);
```
bstewart00, line 50: You might also want to combine this with `insertSnippetAndHreflangTags`

bstewart00: `insertWovnSeoTags`, would be my attempt.

Soichiro725: OK.
How about `InsertSnippetAndLangTags` ?

bstewart00: that will do

test/unit/html/HtmlConverterTest.php
```diff
@@ -238,6 +238,55 @@ public function testInsertSnippetAndHreflangTagsWithFallback()
...
+        foreach ($html_cases as $case) {
+            list($message, $original_html, $expected_html) = $case;
+            $settings = array(
+                'supported_langs' => array('en', 'vi'),
+                'lang_param_name' => 'wovn',
+                'insert_hreflangs' => false
+            );
+            list($store, $headers) = StoreAndHeadersFactory::fromFixture('default', $settings);
+            $converter = new HtmlConverter('UTF-8', $store->settings['project_token'], $store, $headers);
+            $translated_html = $converter->insertHtmlLangAttribute($original_html, 'en');
```
bstewart00, line 284: When you change these test cases to call the public method instead, I would recommend changing the assertion to `translated_html.contains('<html lang="...">`) instead.

src/wovnio/wovnphp/API.php
Soichiro725: done

test/unit/html/HtmlConverterTest.php
Soichiro725: I fixed by `ReflectionMethod` :stuck_out_tongue_closed_eyes:

bstewart00: That's not really good practice either.

Soichiro725: I changed tests.
I think these tests are much better than before.

bstewart00: great. I think you can merge it now.

PR #152, opened 2021-01-05
[Fix]incorrect regex backref

zeyuwu:
In CustomDomainUrlHandler.php, a back-reference is used to replace part of the URL. 

Originally, it was. `"$1${replacement}$3"`. This works fine in most situations (how is this even possible?). But when the URL starts with the number `1`, that `1` will be incorrectly removed.

e.g. `https://123.test.com` -> `https://23.replacement.com` (NG)

Now, the new back-reference is in the `${1}` and `${3}` format.

### DIFF
--- a/src/wovnio/wovnphp/custom_domain/CustomDomainLangUrlHandler.php
+++ b/src/wovnio/wovnphp/custom_domain/CustomDomainLangUrlHandler.php
@@ -20,8 +20,8 @@ class CustomDomainLangUrlHandler
                 "(${currentHostAndPath})". // 2. host and path
                 '((?:/|\?|#).*)?$' . // 3: other
                 '@';
-
-            return preg_replace($regex, "$1${newHostAndPath}$3", $absoluteUrl);
+            $replacement = '${1}' . "${newHostAndPath}" . '${3}';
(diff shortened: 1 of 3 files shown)


src/wovnio/wovnphp/custom_domain/CustomDomainLangUrlHandler.php
```diff
@@ -20,8 +20,8 @@ public static function changeToNewCustomDomainLang($absoluteUrl, $currentCustomD
                 "(${currentHostAndPath})". // 2. host and path
                 '((?:/|\?|#).*)?$' . // 3: other
                 '@';
-
-            return preg_replace($regex, "$1${newHostAndPath}$3", $absoluteUrl);
+            $replacement = '${1}' . "${newHostAndPath}" . '${3}';
```
bstewart00, line 23: does this need to be in a variable though? Sounds like `"${1}${newHostAndPath}${3}"` would be equivalent?

zeyuwu: Ah it doesn't. I left it like that mostly for clarity.

iwai-wovn (approved):

bstewart00 (approved):

PR #153, opened 2021-01-08
[Fix] Change use of SERVER_NAME to HTTP_HOST

zeyuwu:
This PR fixes an issue where the host name of a request cannot be reliably determined when determining the request language.

### Comments
Originally the server environment variable `SERVER_NAME` is used to determine the host name of a request, and this host name is subsequently used to reconstruct a URL for later processing. When a user is using some cgi interfaces, there is a chance that `SERVER_NAME` does not reflect the actual host name.

To fix this, `HTTP_HOST` is used instead, which always reflects the `Host` header of the incoming request. This should be a much more reliable way of determining the host name.

### DIFF
--- a/src/version.php
+++ b/src/version.php
@@ -3,6 +3,6 @@ if (!defined('WOVN_PHP_NAME')) {
     define('WOVN_PHP_NAME', 'WOVN.php');
 }
 if (!defined('WOVN_PHP_VERSION')) {
-    $version = isset($_ENV['WOVN_ENV']) && $_ENV['WOVN_ENV'] === 'development' ? 'VERSION' : '1.8.0';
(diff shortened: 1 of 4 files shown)


iwai-wovn (approved):

bstewart00:
i've always found that usage very weird. Why SERVER_NAME? 🤔

bstewart00 (approved):

zeyuwu:
> Why SERVER_NAME? 🤔

I have no idea. I think when I was writing it I had a brain fart moment. Also it's not helping that in test envs the two are the same.

bstewart00:
That code has been there for years, so it wasn't you ;)

PR #154, opened 2021-01-22
Improve diagnostic tools

zeyuwu:
https://wovnio.atlassian.net/browse/ST-677

### Comments
- added support for wovn.json configuration files
- added server env dump
- added support for custom wovn config file locations
- reduced DIR scan size limit to 1000 items (too huge an HTML file can cause display problems)
- added interpreted config dump

### DIFF
--- a/src/wovnio/wovnphp/Diagnostics.php
+++ b/src/wovnio/wovnphp/Diagnostics.php
@@ -14,6 +14,7 @@ class Diagnostics
 {
     const SALT = '4b73ec6b-a2e1-4ab9-85bc-d92f03d224e5';
     private $store;
+    private $env;
     private $results;
 
     /**
@@ -21,10 +22,11 @@ class Diagnostics
      *
      *  @return void
      */
-    public function __construct($store)
+    public function __construct($store, $env)
     {
-        $this->initResults();
         $this->store = $store;
+        $this->env = $env;
+        $this->initResults();
     }
     
     /**
@@ -77,7 +79,9 @@ class Diagnostics
                 $buffer .= '<table border="solid"><tbody>';
                 foreach ($result as $subItem => &$subResult) {
                     if (is_array($subResult)) {
-                        $subResult = implode(', ', $subResult);
+                        $subResult = json_encode($subResult, JSON_PRETTY_PRINT);
+                    } else {
+                        $subResult = $subResult ? strval($subResult) : '0';
                     }
                     $buffer .= "<tr><td>{$subItem}</td><td>{$subResult}</td></tr>";
                 }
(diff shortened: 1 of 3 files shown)


src/wovnio/wovnphp/Diagnostics.php
```diff
@@ -120,14 +134,22 @@ private function initResults()
...
+                $userSettings = array('Error' => 'Not Found');
+            }
         } else {
-            $userSettings = array('Error' => 'Not Found');
+            $settingFileName = DIRNAME(__FILE__) . '/../../../../wovn.ini';
+            if (file_exists($settingFileName)) {
+                $userSettings = file_get_contents($settingFileName);
+            } else {
+                $userSettings = array('Error' => 'Not Found');
+            }
```
Soichiro725, line 152: I think code like the following is more simple.

```
if (isset($this->env['WOVN_CONFIG'])) {
  $settingFileName = $this->env['WOVN_CONFIG'] ? $this->env['WOVN_CONFIG'] : DIRNAME(__FILE__) . '/../../../../wovn.json';
} else {
  $settingFileName = DIRNAME(__FILE__) . '/../../../../wovn.ini';
}

if (file_exists($settingFileName)) {
  $userSettings = file_get_contents($settingFileName);
} else {
  $userSettings = array('Error' => 'Not Found');
}
```

zeyuwu: Good point! What I wrote was definitely very spaghetti like.

Soichiro725 (approved):
Looks good and this PR works with my environment.

src/wovnio/wovnphp/Diagnostics.php
```diff
@@ -120,12 +134,16 @@ private function initResults()
         );
     }
 
-    private static function getWovnIni()
+    private function getWovnIni()
     {
-        $settingFileName = isset($env['WOVN_CONFIG']) ? $env['WOVN_CONFIG'] : DIRNAME(__FILE__) . '/../../../../wovn.ini';
+        if (isset($this->env['WOVN_CONFIG'])) {
```
aw-leigh, line 139: I'm not very clear on how `WOVN_CONFIG` gets set, does this mean that it'll always use the json file instead of the ini if it's present? If so we should make sure to update the documentation.

zeyuwu: `WOVN_CONFIG` gets set by the upstream server, like in Apache and Nginx. This is passed into PHP as a kind of "global" variable. To set it, in Apache it will be in `.htaccess`, in Nginx it will be `fastcgi_params`

aw-leigh: Got it thanks. Looks good then.

aw-leigh (approved):

PR #155, opened 2021-01-29
Support PHP8

Soichiro725:
https://wovnio.atlassian.net/browse/ST-672
https://wovnio.atlassian.net/browse/ST-673

- Replace deprecated methods in PHP8
- Change test code to work with newer phpunit
- Set up modules by composer with each PHP version

Some syntax change is impossible to share source code for all versions.
So, I used `sed` command to convert code.
I only use `sed` for test code.

### Comments
The followings are detail of changes.
- `: void` syntax is supported from PHP7.1. Without `:void`, some functions like `setUp` in phpunit cause error.
- In shell, regex for digit like `\d` can't be used, so I used [0-9] instead.
- I set up composer by running [composer-setup.php](https://getcomposer.org/installer). But PHP5.3 docker image doesn't have openssl module and have some issuer related to TLS. So, I use zipped composer modules for PHP5.3.
- It is hard to share Dockerfile between test and development. So, I make test.Dockerfile only for test.
- `PHPUnit\Framework\TestCase` doesn't work with PHP5.3. So, I convert test code for PHP5.3.
- From PHP8, `array_key_exists('key', null)` causes fatal error. But under PHP8 is just warning.


src/wovnio/wovnphp/Store.php
```diff
@@ -111,7 +111,7 @@ private function defaultSettings()
             'save_memory_by_sending_wovn_ignore_content' => false,
             'enable_wovn_diagnostics' => false,
             'use_cookie_lang' => false,
-            'wovn_debug_mode' => false
```
Soichiro725, line 114: @aw-leigh I think this param should be `debug_mode`, right?

aw-leigh: Yeah that's probably right.

test/integration/UrlQueryPatternTest.php
```diff
@@ -174,31 +176,32 @@ public function testQueryPatternNotFoundPage()
...
-        'data-wovnio="key=TOKEN&amp;backend=true&amp;currentLang=ja&amp;defaultLang=en&amp;urlPattern=query&amp;langCodeAliases=[]&amp;langParamName=wovn" '.
-        'data-wovnio-info="version=WOVN.php_VERSION" '.
-        'data-wovnio-type="fallback_snippet" '.
-        'async></script>'.
-        '</head>'.
-        '<body>test</body>'.
-        '</html>';
-        $this->assertEquals($content_without_html_swapper, TestUtils::fetchURL('http://localhost/index.html?wovn=ja')->body);
-    }
+    // TODO: Fix me
```
bstewart00, line 179: what's wrong with it?

Soichiro725: WIth PHP8, this test fails. I will check more details.

bstewart00 (approved):

aw-leigh (approved):
Looks good. Nice work!

test/integration/UrlQueryPatternTest.php
Soichiro725: This is because result of `array_key_exists('body', $translation_response)` in API.php.
When `$translation_response` is null, `array_key_exists` with PHP8 cause fatal error.
But under PHP8 just causes warning.

I changed code to check null.
Thank you for pointing out!

PR #156, opened 2021-02-04
Fixed logger default behavior: logger is now off by default unless logging is set.

zeyuwu:
This PR aims to fix a bug where WOVN.php's logger always outputs logs to PHP's system logger, even if no `logging` setting is configured in the config file.

### Comments
**Old behavior:**
The Wovn logger is always set to active upon initialization, with default destination as PHP System Logger.

**New behavior:**
the Wovn logger is only set to active if `logging` setting is configured in the config file.

This PR does not introduce new behaviors and is backward compatible.

### DIFF
--- a/README.en.md
+++ b/README.en.md
@@ -801,7 +801,7 @@ Configures WOVN.php's internal logging. When this section is included in `wovn.j
 
 `max_line_length`: Optional, defaults to `1024`. Lines longer than this setting will be truncated.
 
-`wovn.json` ONLY
+`wovn.json`
 
 ```json
 {
@@ -811,6 +811,14 @@ Configures WOVN.php's internal logging. When this section is included in `wovn.j
     "max_line_length": 5124
   }
 }
+```
+
+`wovn.ini`
+
+```ini
+logging[destination] = "file"
+logging[path] = "/var/logs/error_log.log"
+logging[max_line_length] = 5124
 ```
 
 ## 4. Environment Variable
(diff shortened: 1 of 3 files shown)


Soichiro725:
Can you add document to README.md about log?

zeyuwu:
> Can you add document to README.md about log?

Sure thing. I'm thinking maybe there's some other place that had information on how to configure the logger, because otherwise we wouldn't have people using them. Once I figure out where those information is I'll update those as well.

README.en.md
```diff
@@ -791,6 +791,28 @@ use_cookie_lang = true
...
+
+```json
+{
+  "logging": {
+    "destination": "file",
+    "path": "/var/logs/error_log.log",
+    "max_line_length": 5124
+  }
+}
+```
```
Soichiro725, line 814: wovn.ini can be used like the following format.

```
logging[max_line_length] = 1000
```

Soichiro725:
I'm thinking that we should set default off or not.

zeyuwu:
The code itself is off by default. I don't think users will be too interested in seeing WOVN.php's log unless when there is a problem. If the user want to leave this option on, but not actually see any logs, they can probably make use of the following:

```
"logging": {
  "destination": "file",
  "path": "/dev/null",
  "max_line_length": 1024
}
```

This way, users don't need to remember how to configure the logging, and they can turn logging on whenever there's a need.

Soichiro725:
That's right. I think that's enough.

Soichiro725 (approved):
Looks good

aw-leigh (approved):
Nice catch, look good.

PR #157, opened 2021-02-09
Feature/add support for port numbers

zeyuwu:
https://wovnio.atlassian.net/browse/SUP-619

### Comments
Added support for port numbers in custom domain langs. Port numbers are appended to `host` in parsed urls.

Possible improvement: the code can further be simplified by providing a custom parse_url function.

Custom Domain Lang patterns are now port sensitive.

### DIFF
--- a/README.en.md
+++ b/README.en.md
@@ -308,6 +308,8 @@ NOT SUPPORTED
 }
 ```
 
+*New in 1.8.4*: `url` can contain a port number, e.g. `www.site.com:8081/english`.
+
 For the example above, all request URLs that match `www.site.com/english/*` will be considered as requests in English language.
 All request URLs other than the above that match `www.site.co.jp/*` will be considered as requests in Japanese langauge.
 And, request URLs that match `fr.site.co.jp/*` will be considered as requests in French langauge.
--- a/README.en.md
+++ b/README.en.md
@@ -308,8 +308,6 @@ NOT SUPPORTED
 }
 ```
 
-*New in 1.8.4*: `url` can contain a port number, e.g. `www.site.com:8081/english`.
-
 For the example above, all request URLs that match `www.site.com/english/*` will be considered as requests in English language.
 All request URLs other than the above that match `www.site.co.jp/*` will be considered as requests in Japanese langauge.
 And, request URLs that match `fr.site.co.jp/*` will be considered as requests in French langauge.
--- a/src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
+++ b/src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
@@ -82,7 +82,14 @@ class CustomDomainLangs
     {
         $currentLangDomainLang = $this->getSourceCustomDomainByLang($lang);
         $defaultCustomDomainLang = $this->getCustomDomainLangByLang($defaultLang);
-        return CustomDomainLangUrlHandler::changeToNewCustomDomainLang($physicalUri, $currentLangDomainLang, $defaultCustomDomainLang);
+        $virtualUrl = CustomDomainLangUrlHandler::changeToNewCustomDomainLang($physicalUri, $currentLangDomainLang, $defaultCustomDomainLang);
+        return $this->removePort($virtualUrl);
+    }
+
+    private function removePort($url)
+    {
+        $parsed = parse_url($this->addProtocolIfNeeded($url));
+        return $parsed['scheme'] . '://' . $parsed['host'] . $parsed['path'];
     }
 
     // parse_url needs protocol to parse URL.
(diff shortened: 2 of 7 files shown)


iwai-wovn (approved):

README.en.md
```diff
@@ -308,6 +308,8 @@ NOT SUPPORTED
 }
 ```
 
+*New in 1.8.4*: `url` can contain a port number, e.g. `www.site.com:8081/english`.
```
bstewart00, line 311: I don't think we need to start adding changelogs to the readme?

zeyuwu: Hmm, but it would be very confusing for people if we don't specify a version, wouldn't it?

bstewart00:
Does this affect other patterns than custom domain?

zeyuwu:
No it does not. All changes made are custom domain specific.

bstewart00:
I would probably prefer to see the port number being included in `host` like `host.com:5555`, rather than adding a separate port property. This is similar to how hosts are shown in the browser in `location.host` for example. Would this cause any problems?

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
```diff
@@ -80,7 +82,8 @@ public function computeSourceVirtualUrl($physicalUri, $lang, $defaultLang)
     {
         $currentLangDomainLang = $this->getSourceCustomDomainByLang($lang);
         $defaultCustomDomainLang = $this->getCustomDomainLangByLang($defaultLang);
-        return CustomDomainLangUrlHandler::changeToNewCustomDomainLang($physicalUri, $currentLangDomainLang, $defaultCustomDomainLang);
+        $virtualUrl = CustomDomainLangUrlHandler::changeToNewCustomDomainLang($physicalUri, $currentLangDomainLang, $defaultCustomDomainLang);
+        return $this->removePort($virtualUrl);
```
bstewart00, line 86: What's the significance of removing the port here? According to the issue it says the problem is html swapper is not receiving the correct URL with the port number, which we are removing here.

zeyuwu: We want to send the page to html-swapper as a portless one to avoid having duplicated pages on WOVN's dashboard. This PR is just a temporary fix to make WOVN.php's side of things working, mostly. To fully make things work it would involve changes to foundation.

bstewart00: I don't think we should be making a temporary fix here then. It is much harder to make changes to wovn.php than equalizer.

zeyuwu:
> I would probably prefer to see the port number being included in host like host.com:5555

Yes, when configuring wovn.json the port number is just written there. The only changes we have is that previously when we parse_url a URL we did nothing with the port. Now we included the port in custom domain lang related URL operations.

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
zeyuwu: This wouldn't conflict with our ultimate goal, which is to make libraries port sensitive but not equalizer, right? It's sorta like having frontend does a validation while backend also does its own validation.

README.en.md
bstewart00: I mean each time we have added new features we have been adding to this readme, which won't apply to people that are using previous versions. I just noticed that we mentioned version number before in 
>Starting from version 1.3.0, you must create the configuration file in JSON format if you want to make use of the `custom_domain` URL pattern.
which _maybe_ feels OK since it is a major change, but I think if we start doing this for every minor change (new in 1.8.6, 1.8.7...), it's going to be even more confusing.

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
bstewart00: Equalizer should be the one deciding if port numbers are relevant or not when it comes to whether URLs are equivalent. The library should be providing as much information as possible to make that decision.

README.en.md
zeyuwu: Yeah that's a valid point. But I think we still need to make changes visible and clear to end users. Maybe put this in swapperdocs?

bstewart00: That's what the github releases page is for

src/wovnio/wovnphp/custom_domain/CustomDomainLangs.php
zeyuwu: Ok. Will remove this behavior.

bstewart00:
> > I would probably prefer to see the port number being included in host like host.com:5555
> 
> Yes, when configuring wovn.json the port number is just written there. The only changes we have is that previously when we parse_url a URL we did nothing with the port. Now we included the port in custom domain lang related URL operations.

I understand that, but what i'm saying is a lot of the `if port` checks would be avoided if `parseUrl` simply included the port number as part of the `host` property.

zeyuwu:
Yeah I think I'll include the proposed improvement in this PR then. Will make a custom wrapper for parse_url to automatically include port numbers in the `host`.

bstewart00:
Well hold on, I was under the impression parse_url was our code. If it's a built-in, I wouldn't worry about it then.

zeyuwu:
Yeah it's a built-in. But I'll note this in a `TODO` maybe.

bstewart00 (approved):

bstewart00:
I'm OK with removing the port number from the html swapper request as a temporary measure

PR #161, opened 2021-03-31
Remove unused settings

Soichiro725:
https://wovnio.atlassian.net/browse/ST-837
I removed unused setting parameters.
Some of these params are remained from older WOVNphp.

`wovn_dev_mode` is used to modify html-swapper URL.
But we can use `api_url` instead.


bstewart00 (approved):
Nice

zeyuwu (approved):
Looks good to me.

PR #163, opened 2021-05-04
added new test pages.

zeyuwu:
This PR standardizes the available urls for testing. For PHP pages, they all ends in `.php`.

### Comments
The following standard URLs are made available:
- `/redirecting_page.php` : this will redirect to `/redirection_target.php`
- `/page.php` : a simple html test page

Also, the following .gitignore config doesn't do anything to allow individual files within `/docker/` be tracked...

```
/docker/*
!/docker/apache/*
!/docker/nginx/*
!/docker/HOWTO.md
!/docker/php.ini
```

from git documentaiton:

> It is not possible to re-include a file if a parent directory of that file is excluded. Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined.

I'm not sure why they were added but I'm keeping them for now.

### DIFF
--- a/docker/public/page.php
+++ b/docker/public/page.php
@@ -1,0 +1,33 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta name="description" content="Some description meta." />
+    <meta charset="utf-8">
+    <title>Standard WOVN Test Page</title>
+</head>
+
+<body>
+
+<h1>Website Title (EN)</h1>
+
+<div class="container">
+    <p>This is a sample sentence.</p>
+</div>
+
+<ul>
+    <li>Item Number 1</li>
+    <li>このアイテムは意図的に日本語で書かれています</li>
+</ul>
+
+<div class="photos">
+    <p>This cat photo has alt text.</p>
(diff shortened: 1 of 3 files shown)


Soichiro725 (approved):

docker/public/page.php
```diff
@@ -0,0 +1,33 @@
...
+
+<h1>Website Title (EN)</h1>
+
+<div class="container">
+    <p>This is a sample sentence.</p>
+</div>
+
+<ul>
+    <li>Item Number 1</li>
+    <li>このアイテムは意図的に日本語で書かれています</li>
```
bstewart00, line 20: the source language will be english, so it's a bit strange to have japanese text on it

zeyuwu: The idea was to see if the backend can handle this relatively common scenario properly. But now I think about it it's probably not so much of the library's job and testing it here is a bit weird. What do you think, should we just remove it?

bstewart00: the simpler the better, I think. We won't actually be doing any translation (if we do, we are mocking the html swapper response)

zeyuwu: If that's the case let me update all the page.html to be simpler.

bstewart00 (approved):

PR #164, opened 2021-05-26
add explicit docker network to help integration tests connect to it

bstewart00:
https://wovnio.atlassian.net/browse/ST-699
### Comments
Some changes were required to be able to run these docker environments in circle CI as part of https://github.com/WOVNio/backend_library_integration_tests

It was mainly a problem of networking. We have a separate container for html swapper, and one for wovn.php. They can't talk to each other unless they are running on the same docker network. So I have specified a docker network explicitly. When running locally, this network will be created automatically if it does not exist. If it already exists, then it will use the existing network.

I just chose an arbitrary name like "backend-network" 🤷


aw-leigh (approved):
Nice sleuthing!

Soichiro725 (approved):
Looks good.

PR #165, opened 2021-07-07
[ST-901] Added support for custom requests.

zeyuwu:
https://wovnio.atlassian.net/browse/ST-901

### Comments
This PR is a quality of life PR that improves debugging using the docker environment.

Most notably, this PR adds the ability to send a request to any route prefixed by`/custom_response`, and dictate what the response should be.

You can decide:
- Response headers
- Response HTTP status code
- Response body

You need to include a JSON body with such requests:
```
{
    "response": {
        "body": "foo2",
        "content-type": "text/html",
        "status": 200,
        "headers": { "X-WovnTest": "Foo" }
    }
}
```
Theoretically speaking, content-type is also a header, but it is included in its separate field.

In addition to this universal route, access to normal paths (files) are also supported, as long as they do not contain `custom_response` in their name or path.

**Changes made:**
- No longer copies over configuration files (wovn.ini, .htaccess etc) from the sample files. These files are now directly defined in the web root folder
- a customized wovn_index.php is added to allow custom/universal routes
- a customized .htaccess is added to allow custom/universal routes
- a custom_index.php file is added to handle custom/universal routes

### DIFF
--- a/docker/public/custom_index.php
+++ b/docker/public/custom_index.php
@@ -1,0 +1,12 @@
+<?php
+//require_once('WOVN.php/src/wovn_interceptor.php');
+
+$content = file_get_contents('php://input');
+
+$response_json = json_decode($content, true);
+$response_options = $response_json['response'];
+foreach ($response_options['headers'] as $key => $value) {
+    header("{$key}: {$value}");
+}
+header("Content-Type: {$response_options['content_type']}");
+echo($response_options['body']);
--- a/docker/public/custom_index.php
+++ b/docker/public/custom_index.php
@@ -1,4 +1,6 @@
 <?php
+require_once('WOVN.php/src/wovn_interceptor.php');
+
 $content = file_get_contents('php://input');
 
 $response_json = json_decode($content, true);
--- a/docker/public/custom_index.php
+++ b/docker/public/custom_index.php
@@ -8,5 +8,5 @@ $response_options = $response_json['response'];
 foreach ($response_options['headers'] as $key => $value) {
     header("{$key}: {$value}");
 }
-header("content-type: {$response_options['content_type']}");
+header("Content-Type: {$response_options['content-type']}");
 echo($response_options['body']);
--- a/docker/public/custom_index.php
+++ b/docker/public/custom_index.php
@@ -1,6 +1,4 @@
 <?php
-//require_once('WOVN.php/src/wovn_interceptor.php');
-
 $content = file_get_contents('php://input');
 
 $response_json = json_decode($content, true);
--- a/docker/public/custom_index.php
+++ b/docker/public/custom_index.php
@@ -8,5 +8,5 @@ $response_options = $response_json['response'];
 foreach ($response_options['headers'] as $key => $value) {
     header("{$key}: {$value}");
 }
-header("Content-Type: {$response_options['content_type']}");
+header("content-type: {$response_options['content_type']}");
 echo($response_options['body']);
--- a/docker/public/wovn_index.php
+++ b/docker/public/wovn_index.php
@@ -8,7 +8,7 @@ $parsed_url = parse_url($_SERVER["REQUEST_URI"], PHP_URL_PATH);
 if ($parsed_url) {
     $paths = wovn_helper_detect_paths(dirname(__FILE__), $parsed_url);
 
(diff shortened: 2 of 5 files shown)


docker/public/custom_index.php
```diff
@@ -0,0 +1,12 @@
+<?php
+//require_once('WOVN.php/src/wovn_interceptor.php');
+
+$content = file_get_contents('php://input');
```
zeyuwu, line 4: This is PHP's way of reading the body of a request.

docker/public/wovn_index.php
```diff
@@ -0,0 +1,38 @@
...
+# Enable WOVN.php library
+require_once("WOVN.php/src/wovn_interceptor.php");
+require_once("WOVN.php/src/wovn_helper.php");
+
+$parsed_url = parse_url($_SERVER["REQUEST_URI"], PHP_URL_PATH);
+
+if ($parsed_url) {
+    $paths = wovn_helper_detect_paths(dirname(__FILE__), $parsed_url);
+
+    if (strrpos($_SERVER["REQUEST_URI"], 'custom_response') != false) {
```
zeyuwu, line 11: WOVN's include process depends solely on `REQUEST_URI`, which we are avoiding here because it would result in loading a non-existing file.

bstewart00: `custom_response` requests will never reach this file though right? due to the htaccess rule

zeyuwu: Based on my testing I think the htaccess rewrite rules are cumulative. All requests eventually go through `wovn.index` and get's processed, even for `custom_response`.

bstewart00: Shouldn't be if you add the `[L]` flag to terminate the rewrite rule

zeyuwu: Actually, we don't even need the .htaccess rewrite rule for the custom responses, as the wovn_index.php script is enough to do the redirect.

Soichiro725:
FYI
The following .htaccess works without changing wovn_index.php.

```
RewriteEngine on
RewriteRule ^/?(?:ar|eu|bn|bg|ca|zh-CHS|zh-CHT|da|nl|en|fi|fr|gl|de|el|he|hu|id|it|ja|ko|lv|ms|my|ne|no|fa|pl|pt|ru|es|sw|sv|tl|th|hi|tr|uk|vi|km|ta|si|en-US|zh-Hant-HK)($|/.*$) $1 [L]

RewriteRule ^custom_index.php$ - [L]
RewriteRule ^custom_response/?.* /custom_index.php [L]

# Intercept only static content: html and htm urls
# Warning: do not remove this line or other content could be loaded
RewriteCond %{REQUEST_URI} / [OR]
RewriteCond %{REQUEST_URI} \.(html|htm|shtml|php|php3|phtml)
# Use the wovn_index.php to handle static pages
RewriteRule .? wovn_index.php [L]
```

Soichiro725:
I prefer .htaccess way, because wovn_index.php way has premise that wovn_index.php is used and some code in wovn_index.php like `wovn_helper_include_by_paths` work well.

zeyuwu:
Thank you, this indeed is much better. I've adopted this change, and also made use of `require_once('WOVN.php/src/wovn_interceptor.php');` inside the custom_index.php file to trigger WOVN.

Soichiro725 (approved):
Looks good.

docker/public/custom_index.php
```diff
@@ -0,0 +1,12 @@
...
+require_once('WOVN.php/src/wovn_interceptor.php');
+
+$content = file_get_contents('php://input');
+
+$response_json = json_decode($content, true);
+$response_options = $response_json['response'];
+foreach ($response_options['headers'] as $key => $value) {
+    header("{$key}: {$value}");
+}
+header("Content-Type: {$response_options['content_type']}");
```
bstewart00, line 11: wovnrb version is `content-type`. We should keep the parameter names exactly the same

zeyuwu: Headers are case insensitive by definition, but it's Ok to change this.

bstewart00: i'm talking about the body parameter sent by the browser

zeyuwu: Oops!

bstewart00 (approved):
great, works well

PR #167, opened 2021-09-28
[ST-998] Added support for request ID

zeyuwu:
https://wovnio.atlassian.net/browse/ST-998

### Comments
WOVN.php will assign a "unique" ID to very request upon settings load. The ID is made of `TOKEN` + `current time millisecond component` + `current time second component`.

### DIFF
--- a/src/wovnio/wovnphp/Logger.php
+++ b/src/wovnio/wovnphp/Logger.php
@@ -18,11 +18,12 @@ class Logger
     private $destinationType;
     private $logFilePath;
     private $maxLogLineLength = 1024;
+    private $uniqueId;
 
     public static function get()
     {
         if (self::$logger === null) {
-            self::$logger = new Logger();
+            self::$logger = new Logger('UNKNOWN');
         }
 
         return self::$logger;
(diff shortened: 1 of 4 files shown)


src/wovnio/wovnphp/Logger.php
```diff
@@ -112,15 +119,21 @@ public function debug($message, $context = array())
         $this->log('DEBUG', $message, $context);
     }
 
+    private function getReasonableCurrentTime()
+    {
+        $currentTime = str_replace(' ', '', microtime());
+        $currentTime = str_replace('0.', '', $currentTime);
+        return $currentTime;
+    }
+
```
zeyuwu, line 128: We cannot use anything related to crypto, because some environments might not support them. The timestamp based ID generation is safest.

test/unit/utils/request_handlers/CurlRequestHandlerTest.php
```diff
@@ -16,57 +16,6 @@ protected function tearDown()
...
-
-        $curl_request_handler->expects($this->once())
-            ->method('post')
-            ->with(
-                $this->equalTo($api_url),
-                $this->equalTo($header),
-                $this->equalTo($content),
-                $this->equalTo($timeout)
-            )
-            ->willReturn(array($response, null, null));
```
zeyuwu, line 65: I removed this test because it essentially is just asserting that "the mock will return a variable as it receives it", which it does. But it's pointless.

bstewart00 (approved):
looks good. I like including the token in the id

Soichiro725 (approved):
Looks good.

PR #168, opened 2021-10-14
Do not handle hidden input with no value

zeyuwu:
If a hidden `<input>` doesn't have a `value` specified, we do not need to special handling before sending it to html-swapper, because there is no potentially sensitive information.

### Comments

This avoids the problem of accidentally appending a `value=""` to the `<input>` during our HTML processing.


iwai-wovn (approved):

Soichiro725:
@zeyuwu Update version

PR #169, opened 2021-10-14
feature: [ST-696] add compress_api_requests setting (default true) to allow disabling compression for integration tests.

bstewart00:
https://wovnio.atlassian.net/browse/ST-696
### Comments
add compress_api_requests setting (default true) to allow disabling compression for integration tests.

Also made the apache docker work a bit better, it didn't do the same mappings as nginx.


zeyuwu (approved):
Looks good

PR #170, opened 2021-10-20
[ST-1144] Updated default config and sample config files.

zeyuwu:
https://wovnio.atlassian.net/browse/ST-1144

### Comments
This PR updates the default config and sample config files so that:
- Japanese is the default language
- The default language is also included in the list of `supported_langs`, which is required for correct function of the library.


Soichiro725 (approved):
Looks good.

PR #172, opened 2022-01-26
[ST-1265] Add support for canonical tag

zeyuwu:
https://wovnio.atlassian.net/browse/ST-1265

### Comments
The following changes are made:
- added `translate_canonical_tag`, boolean, default `true`. When `true`,  existing canonical tag URL on the page will be translated.
  - if the canonical tag url is on a different domain than the requested domain, it will not be translated.
  - if this option is `false`, a `translate_canonical_tag: false` will also be sent to html-swapper
- added `api_timeout_search_engine_bots`, integer, default `5`. The number in seconds the html-swapper API call will time out, if the request user-agent indicates it's a search engine bot.


src/wovnio/wovnphp/Headers.php
```diff
@@ -346,4 +346,19 @@ public function shouldRedirect()
...
+    {
+        $canonicalUrl = $this->protocol . '://' . $this->host . $this->pathnameKeepTrailingSlash;
+        if ($this->store->settings['query_canonical_significance'])
+        {
+            $canonicalUrl = $canonicalUrl . '?' . $this->query;
+        }
+        return $canonicalUrl;
+    }
+
+    public function isGoogleBot()
```
bstewart00, line 360: just an early review, but don't name things specific to google bot. You can bet we're going to add more search engines later 😉

zeyuwu: We are not gonna try to support Yahoo, yandex or baidu right.

right?

test/integration/UrlQueryPatternTest.php
```diff
@@ -215,11 +227,13 @@ public function testQueryPatternWithCustomLangAliases()
...
             'supported_langs' => array_keys($langs),
-            'custom_lang_aliases' => $langs
+            'custom_lang_aliases' => $langs,
+            'query_canonical_significance' => true
         ));
 
-        $content_without_html_swapper = '<html lang="en">'.
+        $content_without_html_swapper_custom_en = '<html lang="en">'.
         '<head>'.
+        '<link rel="canonical" href="http://localhost/index.html?wovn=custom_en">'.
```
zeyuwu, line 236: When lang alias is available for default lang, it is incoporated into canonical url.

src/wovnio/wovnphp/Headers.php
```diff
@@ -346,4 +346,18 @@ public function shouldRedirect()
...
+        $canonicalUrl = $this->protocol . '://' . $this->host . $this->pathnameKeepTrailingSlash;
+        if ($this->store->settings['query_canonical_significance'] && !empty($this->query)) {
+            $canonicalUrl = $canonicalUrl . '?' . $this->query;
+        }
+        return $canonicalUrl;
+    }
+
+    public function isSearchEngineBot()
+    {
+        return strpos($this->env['HTTP_USER_AGENT'], 'Googlebot/2.1') !== false;
```
bstewart00, line 361: checking for `/2.1` seems too specific. We don't want to update this every googlebot release.

zeyuwu: I thought about this but there are other bots that we do not want to offer special treatment such as `Googlebot-Image/1.0` and `Googlebot-Video/1.0`...

zeyuwu: I guess we can match `Googlebot/` tho. Should be fine.

test/unit/html/HtmlConverterTest.php
```diff
@@ -196,6 +203,142 @@ public function testInsertSnippetAndLangTagsWithInsertHreflangsFalse()
...
+                'supported_langs' => array('en', 'vi'),
+                'lang_param_name' => 'wovn',
+                'insert_hreflangs' => false,
+                'insert_canonical_tag' => true,
+                'query_canonical_significance' => true
+            );
+            $envs = array(
+                'QUERY_STRING' => '?item=1234',
+                'REQUEST_URI' => '/?item=1234'
+            );
```
bstewart00, line 293: i'm a bit skeptical about needing this structure for a single case, but I would try to move everything constant outside of the foreach ($settings, $envs are the same for all cases)

zeyuwu: Hmm... these `$settings` and `$envs` are already "overrides" of a default set of data. I don't think they are universal between the cases, there are some minor differences.

bstewart00: 😑  do $settings or $envs get modified during the tests?

zeyuwu: They don't get modifed during the same test.

bstewart00 (approved):

bstewart00:
Minor version bump too

Soichiro725:
I think that we should not force inserting canonical tag.
So, correct behavior is that replacing canonical URL to have a target language URL.
Because indexed page will be removed, when target language page has original page's canonical URL.

And, default `insert_canonical_tag` should be false, because canonical tag is not always necessary for multiple languages.

src/wovnio/html/HtmlConverter.php
```diff
@@ -276,10 +276,31 @@ private function insertHreflangTags($html)
         return $this->insertAfterTag($parent_tags, $html, implode('', $hreflangTags));
     }
 
+    private function insertCanonicalTag($html)
+    {
+        if ($this->isNoindexLang($this->headers->requestLang())) {
+            return $html;
+        }
+
+        $canonical_tag = '<link rel="canonical" href="' . $this->buildCanonicalUrl($this->headers->requestLang()) . '">';
```
bstewart00, line 285: `buildCanonicalUrl` is simply delegating to Headers again. Let Headers provide the translated canonical url to begin with. `getCanonicalUrl()` can just return the translated one.

bstewart00 (approved):

src/wovnio/wovnphp/Headers.php
```diff
@@ -346,4 +346,9 @@ public function shouldRedirect()
 
         return $cookieLang && ($requestLang !== $cookieLang) && ($requestLang === $this->store->defaultLang());
     }
+
+    public function isSearchEngineBot()
+    {
+        return strpos($this->env['HTTP_USER_AGENT'], 'Googlebot/') !== false;
```
a-bougouin, line 352: Could we try supporting more search engine:
- `bingbot/`
- `YandexBot/`
- `YandexWebmaster/`
- `DuckDuckBot-Https/`
- `Baiduspider/`

zeyuwu: Initially I wanted to add more support, but I just feel it's not useful to add other search engines because their market share is so small.

src/wovnio/html/HtmlConverter.php
```diff
@@ -276,10 +276,31 @@ private function insertHreflangTags($html)
         return $this->insertAfterTag($parent_tags, $html, implode('', $hreflangTags));
     }
 
+    private function translateCanonicalTag($html)
+    {
+        if ($this->isNoindexLang($this->headers->requestLang())) {
+            return $html;
+        }
+
+        $canonical_tag_regex = "/<link[^>]*rel=\"canonical\"*[^>]*href=\"(.*)\"\>/";
```
bstewart00, line 285: `href=\"(.*)\"` should be `href=\"[^\"]*)\" I think. I'm seeing it match weirdly.

zeyuwu: Yeah you are right, it does look wrong.

test/fixtures/real_html/yelp.html
```diff
@@ -103,7 +103,7 @@
 
 
 
-  <link rel="canonical" href="https://www.yelp.com/search?find_loc=San+Francisco%2C+CA">
+  <link rel="canonical" href="http://my-site.com/">
```
bstewart00, line 106: i'm not expecting these to be modified actually. `"https://www.yelp.com/search?find_loc=San+Francisco%2C+CA"` shouldn't be classified as having the same host as the server (in the tests) so `Url.addLangCode` should be leaving it alone

zeyuwu: Would such a case be expected: the canonical tag is intentionally set to point to a different domain? I feel that because we left the feature to default to false, we assume when the user turn this feature on, they are certain that all canonical tags are correct and valid (at least is on the same domain).

zeyuwu: To be safe I've added a check to verify the canonical URL is same host as request before translating it.

src/wovnio/html/HtmlConverter.php
```diff
@@ -276,10 +276,31 @@ private function insertHreflangTags($html)
...
+        }
+
+        $canonical_tag_regex = "/(<link[^>]*rel=\"canonical\"*[^>]*href=\")([^\"]*)(\"\>)/";
+        preg_match($canonical_tag_regex, $html, $matches);
+        if (count($matches) < 4) {
+            return $html;
+        }
+        $original_canonical_url = $matches[2];
+        $translated_canonical_url = $this->convertUrlToLanguage($original_canonical_url, $this->headers->requestLang());
+        $replacement = '\1' . $translated_canonical_url . '\3';
```
zeyuwu, line 292: basically the idea is to use 3 capture groups with $2 being the canonical URL we want to replace. `\1` and `\3` is just PHP's way of doing back-reference.

src/wovnio/wovnphp/Headers.php
zeyuwu: I've added these bots to the bot list.

a-bougouin: Thanks

README.en.md
```diff
@@ -838,6 +838,47 @@ logging[path] = "/var/logs/error_log.log"
 logging[max_line_length] = 5124
 ```
 
+#### `translate_canonical_tag`
+Configures if WOVN.php should automatically translate existing canonical tag in the HTML. When set to `true`, WOVN.php
+will translate the canonical URL with the current language code. This setting defaults to `true`.
+
+Only language codes and aliases other than the default language will be included in the canonical URL. When the
+default language has an alias, it will be included in the canonical URL.
```
bstewart00, line 846: This explanation feels unnecessary. The semantics of how lang codes are added isn't really the main part of this setting. Just the fact that we are "translating" canonical urls.

README.en.md
```diff
@@ -838,6 +838,47 @@ logging[path] = "/var/logs/error_log.log"
...
+
+`wovn.ini`
+
+```ini
+translate_canonical_tag = true
+```
+
+#### `api_timeout_search_engine_bots`
+A variation of `api_timeout`, configures the translation API request timeout in seconds if WOVN.php has detected a request
+to be coming from a search engine bot. Currently, only GoogleBot is supported.
```
bstewart00, line 864: the googlebot part needs updating

README.en.md
bstewart00: also there's a table with "optional parameters" you should update

bstewart00: Tentative wording i'm using for wovn.cs:
```
Configures if WOVN.cs should automatically translate existing canonical tag in the HTML. When set to `true`, WOVN.cs
will translate the canonical URL with the current language code according to your `url_pattern` setting. This setting defaults to `true`.

Example:
 `<link rel="canonical" href="/page.html">` may be translated to `<link rel="canonical" href="/en/page.html">` depending on your URL pattern.

```XML
  <wovn>
    <add key="translate_canonical_tag" value="true" />
    ...
  </wovn>
```
```

File: README.en.md (hunk 11 above)
Comment (bstewart00) [reply README.en.md:864]: Also we should add this documentation directly after `api_timeout` since they are related

File: src/wovnio/html/HtmlConverter.php [hunk 12]
```diff
@@ -276,10 +276,31 @@ private function insertHreflangTags($html)
         return $this->insertAfterTag($parent_tags, $html, implode('', $hreflangTags));
     }
 
+    private function translateCanonicalTag($html)
+    {
+        if ($this->isNoindexLang($this->headers->requestLang())) {
+            return $html;
+        }
+
+        $canonical_tag_regex = "/(<link[^>]*rel=\"canonical\"[^>]*href=\")([^\"]*)(\"[^>]*>)/";
```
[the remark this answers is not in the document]
zeyuwu: Hmm that's true.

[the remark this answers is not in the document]
zeyuwu: I have added a workaround to account for this situation. I think we still need to check for the existance for both `rel="canonical"` and `href="XXXX"` to know a tag is indeed a canonical tag, so I kept the check for both.

Soichiro725 (approved):
Looks good.

a-bougouin (approved):

PR #175, opened 2022-03-09
[ST-1311] Added support for dynamic loading search engine bot request

zeyuwu:
https://wovnio.atlassian.net/browse/ST-1311

### Comments

- send user-agent to html-swapper
- add a time-stamp to search engine bot requests that changes every 20 minutes

### DIFF
--- a/scripts/run_phpunit_tests.sh
+++ b/scripts/run_phpunit_tests.sh
@@ -62,7 +62,7 @@ docker exec ${APACHE_CONTAINER_ID} \
 # Run unit test
 if [[ "${DOCKER_IMAGE}" =~ ^php:7.*$ ]]; then
     docker exec ${APACHE_CONTAINER_ID} \
-        /bin/bash -c "cd ${WORK_DIR}; set -e; phpdbg -qrr vendor/bin/phpunit --log-junit ${UNITTEST_REPORT_DIR}/results.xml -d memory_limit=1024M --coverage-html ${UNITTEST_REPORT_DIR}/coverage-report"
+        /bin/bash -c "cd ${WORK_DIR}; set -e; vendor/bin/phpunit --log-junit ${UNITTEST_REPORT_DIR}/results.xml -d memory_limit=1024M --coverage-html ${UNITTEST_REPORT_DIR}/coverage-report"
 else
     docker exec ${APACHE_CONTAINER_ID} \
         /bin/bash -c "cd ${WORK_DIR}; set -e; vendor/bin/phpunit --log-junit ${UNITTEST_REPORT_DIR}/results.xml"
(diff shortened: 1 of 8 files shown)


aw-leigh (commented):
Nice work

scripts/run_phpunit_tests.sh
```diff
@@ -62,7 +62,7 @@ docker exec ${APACHE_CONTAINER_ID} \
 # Run unit test
 if [[ "${DOCKER_IMAGE}" =~ ^php:7.*$ ]]; then
     docker exec ${APACHE_CONTAINER_ID} \
-        /bin/bash -c "cd ${WORK_DIR}; set -e; phpdbg -qrr vendor/bin/phpunit --log-junit ${UNITTEST_REPORT_DIR}/results.xml -d memory_limit=1024M --coverage-html ${UNITTEST_REPORT_DIR}/coverage-report"
+        /bin/bash -c "cd ${WORK_DIR}; set -e; vendor/bin/phpunit --log-junit ${UNITTEST_REPORT_DIR}/results.xml -d memory_limit=1024M --coverage-html ${UNITTEST_REPORT_DIR}/coverage-report"
```
Soichiro725, line 65: `phpdbg -qrr` is for output coverage report. So, we can remove it.
Thanks!

src/wovnio/wovnphp/API.php
```diff
@@ -19,17 +21,21 @@ public static function url($store, $headers, $original_content, $request_options
         ksort($store->settings);
         $settings_hash = md5(serialize($store->settings));
         $cache_key_string = "(token=$token&settings_hash=$settings_hash&body_hash=$body_hash&path=$path&lang=$lang)";
+
         if ($request_options->getCacheDisableMode() || $request_options->getDebugMode()) {
```
Soichiro725, line 25: Just an idea, but how about move `$headers->isSearchEngineBot()` to here like the following.
```
if ($request_options->getCacheDisableMode() || $request_options->getDebugMode() || $headers->isSearchEngineBot()) {
```

Then, we can remove `getSearchEngineBotTimeStamp` and make source code simple.

zeyuwu: Hmm. I separated them because the two timestamps are different. Not only are they two different representations, (one is unix epoch type, another is ISO 8601), but they have different underlining "time". The timestamp for debug mode and cache disable mode is updated for every new request, but the timestamp for search engine is updated every 20 minutes.

Soichiro725 (approved):
Looks good.

bstewart00 (approved):

bstewart00:
We didn't push this feature any further in two years, so i'm going to close this

PR #176, opened 2022-04-01
[ST-1349] Stop setting global timezone to UTC by default.

zeyuwu:
https://wovnio.atlassian.net/browse/ST-1349

### Comments
Stop setting the global timezone setting to UTC.

### DIFF
--- a/src/wovnio/wovnphp/Logger.php
+++ b/src/wovnio/wovnphp/Logger.php
@@ -37,7 +37,7 @@ class Logger
     public function __construct($token, $quiet = true, $prefix = 'WOVN.php')
     {
         $this->uniqueId = $token . $this->getReasonableCurrentTime();
-        date_default_timezone_set('UTC');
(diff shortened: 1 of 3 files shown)


src/wovnio/wovnphp/Logger.php
```diff
@@ -37,7 +37,7 @@ public static function set($logger)
     public function __construct($token, $quiet = true, $prefix = 'WOVN.php')
     {
         $this->uniqueId = $token . $this->getReasonableCurrentTime();
-        date_default_timezone_set('UTC');
+        @date_default_timezone_set(@date_default_timezone_get());
```
zeyuwu, line 40: date_default_timezone_get() will return any user-set timezones, and if none is available it will fallback to UTC (but will also emit a warning).

details: https://www.php.net/manual/en/function.date-default-timezone-get.php

iwai-wovn (approved):

bstewart00 (approved):

PR #178, opened 2022-04-20
[ST-1340] Show more detail of file_get_contents error

Soichiro725:
https://wovnio.atlassian.net/browse/ST-1340

- Add more detail of error by using `error_get_last()` to `X-Wovn-Error` header.
  Header will have `([error-type] - [error-in-response-for-html-swapper])`.
  e.g. `X-Wovn-Error: [fgc] Request failed (2 - HTTP\\/1.1 500 Internal Server Error)`
  - error-type: error type code (https://www.php.net/manual/en/errorfunc.constants.php)
  - error-in-response-for-html-swapper: error message in `$http_response_header[0]` (https://www.php.net/manual/en/reserved.variables.httpresponseheader.php)
- Log error message in `error_get_last()`.

### DIFF
--- a/src/wovnio/utils/request_handlers/FileGetContentsRequestHandler.php
+++ b/src/wovnio/utils/request_handlers/FileGetContentsRequestHandler.php
@@ -36,11 +36,10 @@ class FileGetContentsRequestHandler extends AbstractRequestHandler
         list($response, $response_headers) = $this->fileGetContents($url, $http_context);
 
         if ($response === false) {
-            preg_match('{HTTP\/\S*\s(\d{3})}', $response_headers[0], $match);
+            $error_type = error_get_last() ? error_get_last()['type'] : '';
+            $error_in_response = $response_headers[0];
 
-            $http_error_code = $match[1];
-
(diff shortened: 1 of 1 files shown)


aw-leigh (approved):
Nice work.

src/wovnio/utils/request_handlers/FileGetContentsRequestHandler.php
```diff
@@ -36,11 +37,15 @@ protected function post($url, $request_headers, $data, $timeout)
...
         if ($response === false) {
-            preg_match('{HTTP\/\S*\s(\d{3})}', $response_headers[0], $match);
+            $error_type = '';
+            if ($last_error = error_get_last()) {
+                $error_type = $last_error['type'];
+                Logger::get()->error($last_error['message']);
+            }
 
-            $http_error_code = $match[1];
+            $error_in_response = $response_headers[0];
```
zeyuwu, line 46: I googled a bit and it seems that `$http_response_header` can be unset in rare cases (in the case of HTTP 100 for example). So an `@` to suppress any potential warning message would be good.

zeyuwu (approved):
Looks good.

PR #179, opened 2022-05-13
Executes memory free only for PHP5.

zeyuwu:
The SimpleHtmlDOM fork we use suffers from segfaults if the number of nodes in the DOM is high. The segfaults are caused by the `$var -> null;` assignments. Since those are only need for PHP5, a version check is implemented so that this does not affect users of other versions of PHP.


iwai-wovn (approved):

Soichiro725 (approved):
LGTM

aw-leigh (approved):

PR #180, opened 2022-05-31
fix: [ST-1393] implement limited scope default lang alias spec

bstewart00:
https://wovnio.atlassian.net/browse/ST-1393
### Comments
WOVN.php wasn't following our spec exactly for default lang alias.
```
url: http://site.com/page.php (no alias)
removeLang(/page.php, 'ja')
->   $no_lang_uri = Url::removeLangCode($uri, $lang_code, $this->store, $this);
-> /page.php (nothing was removed)
->           return Url::addLangCode($no_lang_uri, $this->store, $default_lang, $this);
-> /japanese/page.php (oops, we just added a lang code we don't want)
```

If you are using default lang alias, the source language pages are assumed to only be accessible via the language alias.
e.g. `/japanese/page.php` is a source language page for Japanese. When a URL has no lang code at all, WOVN should not process the response at all. This is special behavior only for default lang alias  (normally we do handle all urls by default).

Old spec: 
```
path pattern:
/page.php: loads /default_alias/page.php on disk (processed by wovn)
/default_alias/page.php: source lang page. (processed by wovn)
/fr/page.php -> translated version of /default_alias/page.php

subdomain pattern:
site.com: Processed by wovn
default_alias.site.com: Processed by wovn
```

New spec:
```
path pattern:
/page.php : loads /page.php on disk (NOT processed by wovn because it does not include default alias)
/default_alias/page.php: (processed by wovn)
/fr/page.php: (translated version of /default_alias/page.php)

subdomain pattern:
site.com: NOT processed by wovn because it does not include the default alias
default_alias.site.com: Processed by wovn
```


src/wovnio/wovnphp/Headers.php
```diff
@@ -313,19 +313,44 @@ public function removeLang($uri, $lang = null)
 
         $lang_code = $this->store->convertToCustomLangCode($lang);
         $default_lang = $this->store->settings['default_lang'];
-        if ($this->store->hasDefaultLangAlias()) {
+        if ($lang_code && $this->store->hasDefaultLangAlias()) {
```
bstewart00, line 316: This part is a little awkward just because of the way custom lang aliases is implemented. We "remove lang code" followed by "add default lang code".

The goal of `removeLang` is to change `/fr/page.php` into `/default_alias/page.php`. It does not need to do anything when given `/page.php` (where `$lang_code` is empty string).

The else block calls `Url::removeLangCode` which exits early if the lang code provided is empty. This isn't clean but I won't change it here.

src/wovn_interceptor.php
```diff
@@ -68,6 +68,10 @@
         }
         $headers->responseOut();
 
+        if (!$headers->canProcessResponse()) {
+            return $buffer;
+        }
```
bstewart00, line 73: This stops WOVN.php from adding hreflangs and modifying the response. When you are using a default lang alias with path pattern, `/page.php` should not be handled at all.

iwai-wovn (approved):

aw-leigh (approved):

Soichiro725:
I understand about like the following.

Conditions:
- Original lang English / Url pattern `Path` / Target lang Japanese / custom_lang_aliases[en] = en
- Directory
```
Public root
  ┣ en
  ┃ ┗ page.html (B)
  ┗ page.html (A)
```

Old Spec:
- Request example.com/page.html => B
- Request example.com/en/page.html => B
- Request example.com/ja/page.html => B

New Spec:
- Request example.com/page.html => A (WOVN.php doesn't change anything.So, no snippet )
- Request example.com/en/page.html => B
- Request example.com/ja/page.html => B

src/wovnio/wovnphp/Headers.php
```diff
@@ -313,19 +313,44 @@ public function removeLang($uri, $lang = null)
...
+    {
+        # To process a response means to add snippet/hreflangs and translate
+
+        // query pattern doesn't affect which source page is requested
+        // custom domain pattern isn't affected by lang aliases
+        $defaultLangAliasHasLimitedScope = $this->store->settings['url_pattern_name'] == 'path'
+                                        || $this->store->settings['url_pattern_name'] == 'subdomain';
+        if ($defaultLangAliasHasLimitedScope) {
+            $urlLanguage = $this->urlLanguage();
+            $urlLanguageIsEmpty = !$urlLanguage || strlen($urlLanguage)==0;
```
Soichiro725, line 334: Do you know which case we need `strlen($urlLanguage)==0` ?

Soichiro725: Reference
https://leben.mobi/blog/php_isset_etc/php/

bstewart00: I just copied this check from somewhere. I agree it isn't useful but not sure why it is there.

bstewart00: https://github.com/WOVNio/wovnphp/blame/master/src/wovnio/wovnphp/Headers.php#L148 I don't think we'll know because it was from the first commit of the old wovnphp

PR #181, opened 2022-06-06
[SUP-1505] Fix redirect loop with custom domain + source pattern.

bstewart00:
https://wovnio.atlassian.net/browse/SUP-1505?focusedCommentId=49454
### Comments
This can happen with custom domain + source pattern
```
custom_domain_langs: {
   en: { url: "site.com/global/" },
   fr: { url: "site.com/french/", source: "site.com/french/ },
}
```
`site.com/french/page.php` is a different file from `site.com/global/page.php` on disk, and the customer may send a redirect between them.

```
1. Request: https://site.com/french/page.php Lang code: fr
2. Customer redirect: https://site.com/global/page.php
3. WOVN tries to convert redirect to "fr" to maintain language: https://site.com/french/page.php
Redirect loop.
```

I want to try and prevent this by checking after step 3, did we convert the redirect to the same URL as the original request? If so, don't change the location.

### DIFF
--- a/src/version.php
+++ b/src/version.php
@@ -3,6 +3,6 @@ if (!defined('WOVN_PHP_NAME')) {
     define('WOVN_PHP_NAME', 'WOVN.php');
 }
 if (!defined('WOVN_PHP_VERSION')) {
-    $version = isset($_ENV['WOVN_ENV']) && $_ENV['WOVN_ENV'] === 'development' ? 'VERSION' : '1.14.0';
+    $version = isset($_ENV['WOVN_ENV']) && $_ENV['WOVN_ENV'] === 'development' ? 'VERSION' : '1.14.2';
     define('WOVN_PHP_VERSION', $version);
 }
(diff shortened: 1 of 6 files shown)


aw-leigh (approved):
Very thorough tests

src/version.php
```diff
@@ -3,6 +3,6 @@
     define('WOVN_PHP_NAME', 'WOVN.php');
 }
 if (!defined('WOVN_PHP_VERSION')) {
-    $version = isset($_ENV['WOVN_ENV']) && $_ENV['WOVN_ENV'] === 'development' ? 'VERSION' : '1.14.0';
+    $version = isset($_ENV['WOVN_ENV']) && $_ENV['WOVN_ENV'] === 'development' ? 'VERSION' : '1.14.2';
```
Soichiro725, line 6: How about change to `1.14.1` and release once?

Soichiro725 (approved):
Looks good.

PR #184, opened 2022-07-08
remove mapping of sample config files

bstewart00:
### Comments
It was confusing to have `docker/public/wovn.ini` when it wasn't used due to the docker mapping overriding it with `wovn.ini.sample`. I accidentally removed the changes from https://github.com/WOVNio/WOVN.php/pull/165


docker/public/wovn.json
```diff
@@ -0,0 +1,11 @@
+{
+    "project_token": "TOKEN",
```
suplo, line 2: I just thought it would be nicer if we could `sed` the TOKEN with provided token via `--build-arg`.

suplo (approved):

docker/public/wovn.json
bstewart00: interesting idea. Could be useful but most of the time you'll change more than just token and it would be awkward to specify everything via build args imo

iwai-wovn (approved):

Soichiro725 (approved):
Looks good.(In my local, I have always removed these files from docker-compose file)

PR #188, opened 2022-11-18
fix: [ST-1732] Change replace marker for title tag

Soichiro725:
https://wovnio.atlassian.net/browse/ST-1732

The flow of issue is:
1. Customer set `wovn-ignore` attribute for `title` tag. e.g. `<title wovn-ignore>こんにちは</title>`
2. WOVN.php add replace marker for ignored element. e.g. `<title wovn-ignore><!-- __wovn-backend-ignored-key-0 --></title>`
3. html-swapper escapes. e.g. `<title wovn-ignore="">&lt;!-- __wovn-backend-ignored-key-0 --&gt;</title>` 
4. WOVN.php tris to revert it with original value, but `<!-- __wovn-backend-ignored-key-0 -->` doesn't match with `&lt;!-- __wovn-backend-ignored-key-0 --&gt;` and it would remain as marker.

Escape in html-swapper is not wrong, because title tag is just text, so title tag should not include comment node.
https://html.spec.whatwg.org/multipage/semantics.html#the-title-element

This PR changes marker for title tag to not have character which is escaped by html-swapper.
With this PR, HTML after processed replace marker would be like `<title wovn-ignore>__wovn-backend-ignored-key-0</title>`.

### DIFF
--- a/docker/HOWTO.md
+++ b/docker/HOWTO.md
@@ -36,7 +36,10 @@ make clean
 
 ### Debug by XDebug
 #### XDebug configuration for VSCode
-You can use the following setting for `launch.json`.  
+Install `PHP Debug` extension from your Extensions menu in VSCode.  
+Make it possible to load `zend_extension` in `docker/php.ini`.  
+
+You can use the following setting for `.vscode/launch.json`.  
 After start debugging, you can break by break points.  
 ```
 {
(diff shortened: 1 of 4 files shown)


aw-leigh (approved):
Looks like it should work, just needs a test.

iwai-wovn (approved):

bstewart00 (approved):

PR #192, opened 2023-02-06
fix: [ST-1828] wovn_helper now includes files in the global scope

bstewart00:
https://wovnio.atlassian.net/browse/ST-1828
### Comments

Doing `include` inside of a function means the included files inherits the scope of the function. Customers may be declaring global variables inside of the included php file, but when included by `wovn_helper`, they become variables of our function scope, not global scope.

I changed the code to `include` at the top level. This means we have to be careful about variable naming since everything is in the global scope. I renamed all variables and functions to have a `wovn` prefix.

Refactor: I re-arranged the SSI code to be based on a `$wovn_use_ssi` variable rather than have commented out code that the user must uncomment. Because of the changes to fix this bug, it becomes confusing to explain which lines need to be commented out for SSI. It is simpler to have a boolean variable they can turn on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ test/integration/test/integration/wovn_index_sample_workspace
 !/docker/public/wovn.json
 !/docker/public/testdir/*
 !/docker/public/testpage.html
+!/docker/public/ssi_*.html
 !/docker/public/testpage-redirect-origin.php
 !/docker/public/index.html
 !/docker/apache.yml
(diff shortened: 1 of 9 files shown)


bstewart00:
tested that SSI still works, and that global variables defined in the customer PHP files are still global. I don't think it is possible to write a genuine test for the global scope issue.

aw-leigh (approved):

alexisgeoffrey (approved):

iwai-wovn (approved):

PR #195, opened 2023-06-20
feature: [ST-2134] send original request URL to html swapper

bstewart00:
https://wovnio.atlassian.net/browse/ST-2134
https://github.com/WOVNio/equalizer/pull/55201
https://github.com/WOVNio/html-swapper/pull/935

html swapper needs the browser url so it can pass it to equalizer for crawling purposes

### DIFF
--- a/src/wovnio/wovnphp/API.php
+++ b/src/wovnio/wovnphp/API.php
@@ -50,6 +50,7 @@ class API
         $computedUrl = self::getUriRepresentation($headers->urlKeepTrailingSlash, $store, $headers->requestLang());
         $data = array(
             'url' => $computedUrl,  // rewrite URL to use source lang's "virtual" url.
+            'browser_url' => $headers->originalUrl,
             'token' => $token,
             'lang_code' => $headers->requestLang(),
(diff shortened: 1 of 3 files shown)


bstewart00:
I noticed that `$headers->originalUrl` doesn't include the query, which would be a bug. It would only be an issue if the project has the query feature enabled, so for now I will leave it.

src/wovnio/wovnphp/API.php
```diff
@@ -50,6 +50,7 @@ public static function translate($store, $headers, $original_content, $request_o
         $computedUrl = self::getUriRepresentation($headers->urlKeepTrailingSlash, $store, $headers->requestLang());
         $data = array(
             'url' => $computedUrl,  // rewrite URL to use source lang's "virtual" url.
+            'browser_url' => $headers->originalUrl,
```
Soichiro725, line 53: `$headers->originalUrl` meas requested URL which can include lang code. So, it is same URL which is shown in browser. (I feel confused about what is original.)

I will approve.

Soichiro725 (approved):

bryanzerrudo (approved):

PR #197, opened 2023-06-30
test: Add tests to check multiple hidden inputs

bryanzerrudo:

Import tests from WOVNio/wovnjava#229.

### DIFF
--- a/test/unit/html/HtmlConverterTest.php
+++ b/test/unit/html/HtmlConverterTest.php
@@ -1118,24 +1118,21 @@ bye
     {
         $original_html = '<html><body><head></head>';
         $expected_converted_html = '<html><body><head></head>';
-        $expected_reverted_html = '<html><body><head></head>';
 
         for ($i = 0; $i < 50; $i++) {
             $original_html .= "<input type=\"hidden\" name=\"field_{$i}\" value=\"text-{$i}\">";
             $expected_converted_html .= "<input type=\"hidden\" name=\"field_{$i}\" value=\"__wovn-backend-ignored-key-{$i}\">";
-            $expected_reverted_html .= "<input type=\"hidden\" name=\"field_{$i}\" value=\"text-{$i}\">";
         }
 
         $original_html .= '</body></html>';
         $expected_converted_html .= '</body></html>';
-        $expected_reverted_html .= '</body></html>';
 
         list($store, $headers) = StoreAndHeadersFactory::fromFixture('default');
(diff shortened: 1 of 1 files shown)


test/unit/html/HtmlConverterTest.php
```diff
@@ -1114,6 +1114,30 @@ public function testConvertToAppropriateBodyForApiWithMultipleComment()
...
     }
 
+    public function testConvertToAppropriateBodyForApiWithFormHavingMultipleInputs()
+    {
+        $original_html = '<html><body><head></head>';
+        $expected_converted_html = '<html><body><head></head>';
+        $expected_reverted_html = '<html><body><head></head>';
+
+        for ($i = 0; $i < 50; $i++) {
+            $original_html .= "<input type=\"hidden\" name=\"field_{$i}\" value=\"text-{$i}\">";
```
bryanzerrudo, line 1124: Note: `value` here cannot be `'0'`, otherwise tests would fail, e.g.:
```
--- Expected
+++ Actual
@@ @@
-'<html><body><head></head><input type="hidden" name="field_0" value="__wovn-backend-ignored-key-0"> // ...
+'<html><body><head></head><input type="hidden" name="field_0" value="0"> // ...
```

This is because of:
https://github.com/WOVNio/WOVN.php/blob/77ef83da18b2d6f35a9bd6a065b23c3016e83c27/src/wovnio/html/HtmlConverter.php#L418

in which `!$originalText` returns truthy value.

bstewart00: probably worth fixing since numbers are likely to be used for hidden inputs

Soichiro725 (approved):

test/unit/html/HtmlConverterTest.php
bryanzerrudo: Let's do it on a separate PR?

bryanzerrudo: Seems like there's an issue with `getAttribute` (ref. 76efbbd3cc6b2166a587340bf34b55ae995b20a6):
> 2) Wovnio\Wovnphp\Tests\Unit\ModifiedVendor\SimpleHtmlDomTest::testGetAttribute
Failed asserting that false matches expected '0'.

bstewart00: if it's not trivial we can do it in a separate PR. But I suspect we can just remove that check entirely and always call strpos. checking for false doesn't seem right.

bstewart00 (approved):

PR #199, opened 2023-07-10
fix: [ST-2184] Hidden input with "0" value is not ignored

bryanzerrudo:

Fixes logic where `0` is not ignored and sent to html-swapper as-is.

Before (text sent to html-swapper):
```
<input type="hidden" value="0">
```

After:
```
<input type="hidden" value="__wovn-backend-ignored-key-${SOME_INCREMENTAL_KEY}">
```

### Comments

PHP considers `'0'` string as `false` when converted to boolean (ref. [docs](https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting)).

### DIFF
--- a/src/wovnio/html/HtmlConverter.php
+++ b/src/wovnio/html/HtmlConverter.php
@@ -415,7 +415,7 @@ class HtmlConverter
 
         if (strtolower($node->tag) === 'input' && strtolower($node->getAttribute('type')) == 'hidden') {
             $originalText = $node->getAttribute('value');
-            if ($originalText === false || strpos($originalText, HtmlReplaceMarker::$keyPrefix) !== false) {
+            if ((!$originalText && $originalText !== '0') || strpos($originalText, HtmlReplaceMarker::$keyPrefix) !== false) {
                 return;
             }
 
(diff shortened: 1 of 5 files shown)


src/wovnio/html/HtmlConverter.php
```diff
@@ -415,7 +415,7 @@ public function _removeForm($node, $marker)
 
         if (strtolower($node->tag) === 'input' && strtolower($node->getAttribute('type')) == 'hidden') {
             $originalText = $node->getAttribute('value');
-            if (!$originalText || strpos($originalText, HtmlReplaceMarker::$keyPrefix) !== false) {
+            if ((!$originalText && $originalText !== '0') || strpos($originalText, HtmlReplaceMarker::$keyPrefix) !== false) {
```
Soichiro725, line 418: How about check with string length like `strlen($originalText) > 0` ?

bryanzerrudo: Agree, that's better. Fixed in e9831e79e8b11923bae7cfece62da0292451c9ae.

Soichiro725 (approved):

src/wovnio/html/HtmlConverter.php
bstewart00: I don't really like this either. What does `getAttribute` return when the attribute doesn't exist?

bryanzerrudo: Checking explicitly for `false` was what I originally had in 76efbbd3cc6b2166a587340bf34b55ae995b20a6. But I changed it to be more permissive to other falsy values. In the end, I don't think that's correct since what we want is to check when the attribute doesn't exist, as you said.

Fixed in 566846b4b91384ec31e062b455d510a98e3a2eba.

bstewart00 (approved):

PR #200, opened 2023-08-10
feature: [ST-2232] support meta refresh tags

bstewart00:
https://wovnio.atlassian.net/browse/ST-2232
https://github.com/WOVNio/html-swapper/pull/968
### Comments
Some customers are using refresh meta tags to trigger refreshes. We want to translate this redirect URL similar to a normal redirect.

`<meta http-equiv="refresh" content="0;URL=/international/index.html#anchor02">`

### DIFF
--- a/src/wovnio/html/HtmlConverter.php
+++ b/src/wovnio/html/HtmlConverter.php
@@ -378,7 +378,7 @@ class HtmlConverter
      * Note: Because php5.3 doesn't allow calling private method inside anonymous function,
      * Use `_` prefix to imply `private`
      */
-    private function _translateMetaTagLink($meta_node)
+    public function _translateMetaTagLink($meta_node)
     {
         $httpEquiv = $meta_node->getAttribute('http-equiv');
         $metaContent = $meta_node->getAttribute('content');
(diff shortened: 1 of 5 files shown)


src/wovnio/html/HtmlConverter.php
```diff
@@ -370,6 +374,29 @@ private function putReplaceMarker($element, $marker)
...
+     * Note: Because php5.3 doesn't allow calling private method inside anonymous function,
+     * Use `_` prefix to imply `private`
+     */
+    public function _translateMetaTagLink($meta_node)
+    {
+        $httpEquiv = $meta_node->getAttribute('http-equiv');
+        $metaContent = $meta_node->getAttribute('content');
+
+        if ($httpEquiv === 'refresh') {
+            $splitMetaContent = preg_split('/;(\s*url)=/i', $metaContent, null, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
```
bryanzerrudo, line 387: Should we ignore non-numeric refresh time?
```suggestion
            $splitMetaContent = preg_split('/(\d+);(\s*url)=/i', $metaContent, null, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
```

iwai-wovn (approved):

src/wovnio/html/HtmlConverter.php
bstewart00: for our purposes we don't care, we are only interested in the url part regardless of how the browser processes the directive later

bstewart00:
Looks good (I started this PR but Bryan finished it)

bryanzerrudo (approved):

PR #202, opened 2023-10-23
feature: add outbound_proxy_host/port settings, to support HTTP proxy via curl

bstewart00:
https://wovnio.atlassian.net/browse/ST-2444
### Comments
Support connecting to wovn servers via HTTP proxy. Only supported for curl.

Didn't test with an actual proxy server but curl reported this error so the setting is being applied correctly
```
curl_error($curl_session)
"Failed to connect to proxy.com port 7979: Connection timed out"
```

### DIFF
--- a/src/wovnio/wovnphp/Store.php
+++ b/src/wovnio/wovnphp/Store.php
@@ -265,18 +265,17 @@ class Store
         return $this->settings['compress_api_requests'];
     }
 
-    public function outboundProxy()
+    public function outboundProxyHost()
     {
-        $result = null;
+        return isset($this->settings['outbound_proxy_host'])
+            ? $this->settings['outbound_proxy_host']
+            : null;
+    }
 
-        if (isset($this->settings['outbound_proxy_host'])) {
-            $result = $this->settings['outbound_proxy_host'];
-        }
-        if (isset($this->settings['outbound_proxy_port'])) {
-            $port = $this->settings['outbound_proxy_port'];
-            $result .= ":$port";
-        }
-
-        return $result;
+    public function outboundProxyPort()
+    {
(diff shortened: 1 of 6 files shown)


iwai-wovn (approved):

src/wovnio/wovnphp/Store.php
```diff
@@ -260,4 +264,19 @@ public function compressApiRequests()
...
+
+        if (isset($this->settings['outbound_proxy_host'])) {
+            $result = $this->settings['outbound_proxy_host'];
+        }
+        if (isset($this->settings['outbound_proxy_port'])) {
+            $port = $this->settings['outbound_proxy_port'];
+            $result .= ":$port";
+        }
+
+        return $result;
```
Soichiro725, line 280: Why do we need to split host and port? How about having configuration to be connected.
e.g. `outbound_proxy_host_and_port: site.com:8000`

bryanzerrudo: Seems like there's a `CURLOPT_PROXYPORT`:
https://www.php.net/manual/en/function.curl-setopt.php (under "options")

If we want, we can use that to specify the port.

bstewart00: to be consistent with wovnjava 😎

bstewart00: They do mention this re CURLOPT_PROXYPORT
> The proxy's port number may optionally (but discouraged) be specified with the separate option [CURLOPT_PROXYPORT](https://curl.se/libcurl/c/CURLOPT_PROXYPORT.html).

Soichiro725 (approved):

PR #204, opened 2024-05-27
refactor: [ST-2960] Fix deprecations

bryanzerrudo:

https://wovnio.atlassian.net/browse/ST-2960

Changes:
- Fixes the following deprecations:
  - "Deprecate ${} string interpolation" (see [PHP RFC](https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation))
  - "Deprecate dynamic properties" (see [PHP RFC](https://wiki.php.net/rfc/deprecate_dynamic_properties))
- Adds tests for PHP 8.1 to 8.3

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,27 @@ jobs:
           paths:
             - ./vendor
       - run: ./vendor/bin/phpcs
+  test_php83:
+    <<: *defaults
+    steps:
+      - common-steps
+      - run: scripts/run_phpunit_tests.sh php:8.3-apache
+      - run: scripts/run_integration_tests.sh php:8.3-apache
+      - store-steps
+  test_php82:
+    <<: *defaults
+    steps:
+      - common-steps
+      - run: scripts/run_phpunit_tests.sh php:8.2-apache
+      - run: scripts/run_integration_tests.sh php:8.2-apache
+      - store-steps
+  test_php81:
+    <<: *defaults
+    steps:
+      - common-steps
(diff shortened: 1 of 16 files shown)


.circleci/config.yml
```diff
@@ -20,6 +20,7 @@ commands:
           at: ~/project
       - setup_remote_docker:
           docker_layer_caching: true
+          version: docker25
```
bryanzerrudo, line 23: See also: [Docker docs](https://docs.docker.com/engine/deprecated/#pushing-and-pulling-with-image-manifest-v2-schema-1)

bstewart00 (approved):

bstewart00 (approved):
since production code was affected, bump patch version

bryanzerrudo:
> since production code was affected, bump patch version

I was considering not making a release this time. WDYT?

Soichiro725:
> > since production code was affected, bump patch version
> 
> Since there are no behavior changes, I was considering not making a release this time. WDYT?

I think we should bump version, when we change production code. So, we should make new release in this case.

test/unit/StoreTest.php
```diff
@@ -28,8 +28,6 @@ private function getMockAndRegister($originalClassName, $methods)
         $mockObject = $builder->getMock();
         if (method_exists($this, 'registerMockObject')) {
             $this->registerMockObject($mockObject);
-        } else {
-            $this->mockObjects[] = $mockObject;
```
bryanzerrudo, line 32: Unused variable.

iwai-wovn (approved):

PR #207, opened 2024-11-19
feat: [ST-3448] Add `hreflang_x_default_lang` setting

bryanzerrudo:

https://wovnio.atlassian.net/browse/ST-3448

See `README` changes for setting overview.

### DIFF
--- a/test/unit/html/HtmlConverterTest.php
+++ b/test/unit/html/HtmlConverterTest.php
@@ -200,7 +200,7 @@ class HtmlConverterTest extends TestCase
     {
         $html_cases = array(
             array (
-                'has setting and html has none - is generated',
+                'has x-default setting and html has none - is generated',
 
                 '<html><head></head><body><a>hello</a></body></html>',
 
@@ -220,7 +220,7 @@ class HtmlConverterTest extends TestCase
                 'en'
             ),
             array (
-                'has setting and html has existing - is not modified',
+                'has x-default setting and html has existing - is not modified',
 
                 '<html>' .
                 '<head>' .
@@ -247,7 +247,7 @@ class HtmlConverterTest extends TestCase
                 'en'
             ),
             array (
-                'no setting and html has existing - is not modified',
+                'no x-default setting and html has existing - is not modified',
 
                 '<html>' .
                 '<head>' .


Soichiro725:
I want to see the purpose of this PR in this description not in the link. (Jira ticket of link doesn't have description. Parent ticket has description. But ...)

Soichiro725 (approved):

src/wovnio/html/HtmlConverter.php
```diff
@@ -280,6 +280,16 @@ private function insertHreflangTags($html)
             array_push($hreflangTags, '<link rel="alternate" hreflang="' . Lang::iso6391Normalization($lang_code) . '" href="' . $href . '">');
         }
 
+        if (isset($this->store->settings['hreflang_x_default_lang'])) {
+            $x_default_hreflang_regex = "/<link [^>]*hreflang=[\"']?(x-default)[\"']?(\s[^>]*)?\>/iU";
```
Soichiro725, line 284: I prefer the following regex. I think it works as same way.
```
<link[^>]*hreflang=[\"']?x-default[\"']?[^>]*>
```
- No need to do capture
- `[^>]` can include space
- `*` can be nothing, so no need to add `?`

bryanzerrudo: Fixed in https://github.com/WOVNio/WOVN.php/pull/207/commits/40443223bc9f64743e846c214cbcd596f871688b.

README.en.md
```diff
@@ -932,6 +933,31 @@ outbound_proxy_host = site.com
 outbound_proxy_port = 8080
 ```
 
+#### `hreflang_x_default_lang`
+Configures the language code used to generate link tag with `hreflang="x-default"` attribute.
+
+The language code used must be in `supported_langs`. If the language code is invalid, link tag will not be inserted.
+
+If a link tag with `hreflang="x-default"` attribute exists, this setting does nothing.
```
Soichiro725, line 941: When customer uses `custom_domain` pattern, there is a case we need to overwrite link tag with `hreflang="x-default"` when customer has config to change their default lang URL.

Soichiro725: For now, I agree with current spec.

Soichiro725 (approved):

test/unit/html/HtmlConverterTest.php
```diff
@@ -196,6 +196,100 @@ public function testInsertSnippetAndLangTagsWithInsertHreflangsFalse()
         }
     }
 
+    public function testInsertSnippetAndLangTagsWithHreflangXDefaultLang()
+    {
+        $html_cases = array(
+            array (
+                'has setting and html has none - is generated',
```
bstewart00, line 203:
```suggestion
                'has x-default setting and html has none - is generated',
```

test/unit/html/HtmlConverterTest.php
```diff
@@ -196,6 +196,100 @@ public function testInsertSnippetAndLangTagsWithInsertHreflangsFalse()
...
+                '</head>' .
+                '<body>' .
+                '<a>hello</a>' .
+                '</body>' .
+                '</html>',
+
+                'en'
+            ),
+            array (
+                'has setting and html has existing - is not modified',
```
bstewart00, line 223:
```suggestion
                'has x-default setting and html has existing - is not modified',
```

test/unit/html/HtmlConverterTest.php
```diff
@@ -196,6 +196,100 @@ public function testInsertSnippetAndLangTagsWithInsertHreflangsFalse()
...
+                '</head>' .
+                '<body>' .
+                '<a>hello</a>' .
+                '</body>' .
+                '</html>',
+
+                'en'
+            ),
+            array (
+                'no setting and html has existing - is not modified',
```
bstewart00, line 250:
```suggestion
                'no x-default setting and html has existing - is not modified',
```

bstewart00 (approved):

PR #209, opened 2025-09-25
feature: always emit x-default hreflang (as long as no existing in html). use default lang if no setting

bstewart00:
https://wovnio.atlassian.net/browse/ST-3895
https://github.com/WOVNio/wovnjava/pull/237
https://github.com/WOVNio/Wovn.cs/pull/96
https://github.com/WOVNio/wovnrb/pull/248
https://github.com/WOVNio/WOVN.php/pull/209
https://github.com/WOVNio/proxy/pull/193
https://github.com/WOVNio/wordpress_plugin/pull/130
https://github.com/WOVNio/html-swapper/pull/1179
https://github.com/WOVNio/equalizer/pull/73341

`<link rel="alternate" hreflang="X-Default" href="https://example.com/fr/">` will now _always_ be generated if the customer does not already have one on their page. The href will be based on the setting. If the setting doesn't exist, the default lang will be used.

html swapper will receive the raw setting, without calculating a fallback. Since html swapper can calculate it more accurately based on js_data.
### Comments

New tests:
- no customer x-default, generates using default lang if no setting
- no customer x-default, generates using setting language
- customer has x-default, does not modify
- do not generate if this language is contained in `no_index_langs or no_hreflang_langs`


src/wovnio/html/HtmlConverter.php
```diff
@@ -257,16 +257,7 @@ private function buildParamsStr($params_array)
...
-        if (isset($this->store->settings['supported_langs'])) {
-            if (is_array($this->store->settings['supported_langs'])) {
-                $lang_codes = $this->store->settings['supported_langs'];
-            } else {
-                $lang_codes = array($this->store->settings['supported_langs']);
-            }
-        } else {
-            $lang_codes = array();
-        }
-
```
Soichiro725, line 269: **question:** Existing code looks like trying to force array lang_codes. I feel scared to remove logic for it.

Soichiro725: Ah, you just moved logic.

Soichiro725 (approved):

bryanzerrudo (approved):
**todo:** Version bump!

bryanzerrudo (approved):

PR #213, opened 2026-07-23
fix: undefined $marker in the no-request-handler branch of API::translate

masaru87:
In `API::translate()` (`src/wovnio/wovnphp/API.php`), the branch taken when no request handler is available references an undefined variable:

```php
$request_handler = RequestHandlerFactory::getBestAvailableRequestHandler($store);
if ($request_handler === null) {
    return $marker->revert($converted_html);   // $marker is never defined in this method
}
```

`$marker` is not defined anywhere in `translate()` (it appears only on this line). Every other return in the method uses `$converter->revertMarkers($converted_html)` (the `HtmlConverter` created at the top of the method). So this line is both an undefined variable and the wrong method name.

`getBestAvailableRequestHandler()` returns `null` when neither request handler is available (e.g. the cURL extension is missing and `allow_url_fopen` is off). In that situation, instead of the intended graceful degradation (returning the page untranslated), calling `->revert()` on the undefined `$marker` throws an `\Error`. The surrounding `catch (\Exception $e)` does not catch `\Error`, so it becomes an uncaught fatal — and since WOVN.php wraps the whole page output, every request on such a server returns a 500 rather than the untranslated page.

Fixed by calling `$converter->revertMarkers($converted_html)`, matching the other branches.

### DIFF
--- a/src/version.php
+++ b/src/version.php
@@ -3,6 +3,6 @@ if (!defined('WOVN_PHP_NAME')) {
     define('WOVN_PHP_NAME', 'WOVN.php');
 }
 if (!defined('WOVN_PHP_VERSION')) {
-    $version = isset($_ENV['WOVN_ENV']) && $_ENV['WOVN_ENV'] === 'development' ? 'VERSION' : '1.28.1';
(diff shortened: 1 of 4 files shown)


bstewart00 (approved):
Looks good.

